Issue #20, opened 2025-03-19
Rename "service" classes to "API"


robbieaverill:
We have some "service" classes, e.g. `BaseServiceClient`. These should be named `API` rather than `Service`, to better align with the [developer documentation](https://developer.seequent.com/), where we refer to Evo APIs - for example: Geoscience Object API.

There is a PEP 8 standard that covers the acronym API being uppercase: https://legacy.python.org/dev/peps/pep-0008/#descriptive-naming-styles.

Examples:
* FileAPIClient ✅ 
* BaseServiceClient  -> BaseAPIClient
* ObjectServiceClient -> ObjectAPIClient

PR #24, opened 2025-03-25
Rename classes to use `API` consistently

robbieaverill:
## Description

Many classes were using `Api` in their name. PEP-8 specifies that class names should use capitalized words, and that when using acronyms in CapWords, captilize all the letters of the acronym (e.g. `Api` should be `API`).

Some of the classes in our auto-generated SDK clients are unchanged here, I will work out how to update them separately. We don't consider them part of our public API contract, and don't intend for developers to consume them directly, so I expect we can change them without breaking semver commitments later.

High level summary of changes:
* `ApiConnector` renamed to `APIConnector`
* `EvoApiException` renamed to `EvoAPIException`
* `DiscoveryApiClient` renamed to `DiscoveryAPIClient` (discovery is not auto-generated)
* Regenerated file, object, and workspaces clients
* `*ServiceClient` renamed to `*APIClient`
* `FileApiUpload` and `FileApiDownload` renamed to `FileAPIUpload` and `FileAPIDownload`

Resolves https://github.com/SeequentEvo/evo-python-sdk/issues/20

## Checklist


packages/evo-client-common/src/evo/notebooks/_consts.py
```diff
@@ -0,0 +1,18 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+DEFAULT_DISCOVERY_URL = "https://discover.api.seequent.com"
+DEFAULT_ISSUER_URL = "https://ims.bentley.com"
+DEFAULT_REDIRECT_URL = "http://localhost:3000/signin-oidc"
+DEFAULT_CLIENT_ID = "evopythonsdk"
+DEFAULT_DEVICE_FLOW_CLIENT_ID = "evo-sdk-device-flow"
```
wordsworthc, line 16: These two client IDs are only registered on UAT. Who can we designate to create/maintain a default client ID for prod to be included in this codebase?

packages/evo-client-common/pyproject.toml
```diff
@@ -16,6 +16,12 @@ build-backend = "setuptools.build_meta"
 
 [project.optional-dependencies]
 aiohttp = ["aiohttp"]
+notebooks = [
+    "evo-client-common[aiohttp]",
+    "ipywidgets",
+    "jupyter",
```
wordsworthc, line 22: Possibly worth checking if `jupyter` is actually required by us specifically. It may end up being a dev dependency, as opposed to an SDK dependency

packages/evo-client-common/src/evo/notebooks/_consts.py
robbieaverill: @davidknight-seequent any ideas here?

davidknight-seequent: This is something that will require approval from higher up, plus I just noticed this new warning in the iTwin portal: 
CleanShot 2025-04-04 at 10 42 41@2x

mpj17 (approved):

packages/evo-client-common/src/evo/notebooks/_consts.py
robbieaverill: We will remove the default values for these parameters and make them required. Clients will need to follow the documentation at https://developer.seequent.com/docs/guides/getting-started/apps-and-tokens and create their own client before they'll be able to use Evo, or these notebooks. This is a behavioural change from the internal audience these notebooks were originally built for, but can be solved with some documentation.

PR #30, opened 2025-04-03
Change port binding to be more specific in aio test

robbieaverill:
## Description

This has no impact on the test or usable SDK code, but keeps automatic code checking tools happy.

## Checklist

### DIFF
--- a/packages/evo-client-common/tests/aio/test_transport.py
+++ b/packages/evo-client-common/tests/aio/test_transport.py
@@ -132,7 +132,7 @@ class MockConnector(mock.AsyncMock):
 
 def _get_open_port() -> int:
     s = socket.socket(socket.AF_INET, type=socket.SOCK_STREAM)
-    s.bind(("", 0))
+    s.bind(("127.0.0.1", 4000))
     _, port = s.getsockname()
     s.close()
     return port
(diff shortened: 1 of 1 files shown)


packages/evo-client-common/tests/aio/test_transport.py
```diff
@@ -132,7 +132,7 @@ def klass(cls) -> mock.Mock:
 
 def _get_open_port() -> int:
     s = socket.socket(socket.AF_INET, type=socket.SOCK_STREAM)
-    s.bind(("", 0))
+    s.bind(("127.0.0.1", 4000))
```
wordsworthc, line 135: hard-coding a specific port number could fail if that specific port is already used. is there another way we can use any available port?

robbieaverill: I'm not sure it's worth the extra logic. What if I arbitrarily use port 26384 and assume that won't be in use?

wordsworthc: Was the tooling complaining specifically about binding to any available port? or just the `0.0.0.0` address bind?

robbieaverill: It's the interface actually, not the port. I'll update.

wordsworthc (approved):

l-macdonald (approved):

PR #50, opened 2025-04-29
Add AccessTokenAuthorizer to authorize requests with the provided token

robbieaverill:
## Description

If you already have a token and you simply need to authoriser requests with it, this authorizer skips OAuth processes that other `IAuthorizer` implementations use and adds the access token to the authentication header for the request.

## Checklist

### DIFF
--- a/packages/evo-sdk-common/src/evo/oauth/authorizer.py
+++ b/packages/evo-sdk-common/src/evo/oauth/authorizer.py
@@ -372,9 +372,9 @@ class AccessTokenAuthorizer(IAuthorizer):
         """
         self._access_token = access_token
 
-    def refresh_token(self) -> bool:
+    async def refresh_token(self) -> bool:
         logger.debug("AccessTokenAuthorizer does not support refreshing expired tokens.")
         return False
 
-    def get_default_headers(self) -> HTTPHeaderDict:
+    async def get_default_headers(self) -> HTTPHeaderDict:
         return HTTPHeaderDict({"Authorization": f"Bearer {self._access_token}"})
--- a/packages/evo-sdk-common/src/evo/oauth/authorizer.py
+++ b/packages/evo-sdk-common/src/evo/oauth/authorizer.py
@@ -25,6 +25,7 @@ from .oauth_redirect_handler import OAuthRedirectHandler
 from .oidc import OIDCConnector
 
 __all__ = [
+    "AccessTokenAuthorizer",
     "AuthorizationCodeAuthorizer",
     "ClientCredentialsAuthorizer",
     "DeviceFlowAuthorizer",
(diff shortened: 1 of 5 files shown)


packages/evo-sdk-common/src/evo/oauth/authorizer.py
```diff
@@ -360,3 +361,20 @@ def _validate_configuration(self):
...
+class AccessTokenAuthorizer(IAuthorizer):
+    def __init__(self, access_token: str):
+        """An OAuth authorizer that uses the provided access token to authorize requests.
+
+        This authorizer does not make any attempt to refresh expired tokens. This is handy if you already have an
+        access token and are happy to support refreshing it when it expires outside of this interface.
+        """
+        self._access_token = access_token
+
+    def refresh_token(self) -> bool:
```
wordsworthc, line 375:
```suggestion
    async def refresh_token(self) -> bool:
```

packages/evo-sdk-common/src/evo/oauth/authorizer.py
```diff
@@ -360,3 +361,20 @@ def _validate_configuration(self):
...
+        logger.debug("AccessTokenAuthorizer does not support refreshing expired tokens.")
+        return False
+
+    def get_default_headers(self) -> HTTPHeaderDict:
```
wordsworthc, line 379:
```suggestion
    async def get_default_headers(self) -> HTTPHeaderDict:
```

packages/evo-sdk-common/tests/oauth/test_access_token_authorizer.py
```diff
@@ -0,0 +1,12 @@
...
+
+class TestAccessTokenAuthorizer:
+    def test_get_default_headers(self) -> None:
+        authorizer = AccessTokenAuthorizer(access_token="abc-123")
+        headers = authorizer.get_default_headers()
+        assert headers == {"Authorization": "Bearer abc-123"}
+
+    def test_refresh_token(self) -> None:
+        authorizer = AccessTokenAuthorizer(access_token="abc-123")
+        assert not authorizer.refresh_token()
```
wordsworthc, line 12:
```suggestion
import unittest

from evo.oauth.authorizer import AccessTokenAuthorizer


class TestAccessTokenAuthorizer(unittest.IsolatedAsyncioTestCase):
    def setUp(self) -> None:
        self.authorizer = AccessTokenAuthorizer(access_token="abc-123")
    
    async def test_get_default_headers(self) -> None:
        headers = await self.authorizer.get_default_headers()
        assert headers == {"Authorization": "Bearer abc-123"}

    async def test_refresh_token(self) -> None:
        assert not await self.authorizer.refresh_token()
```

wordsworthc (changes requested):

packages/evo-sdk-common/src/evo/oauth/authorizer.py
robbieaverill: Interesting that Python doesn't complain about this. Is it not a problem, or is Python's compiler not mature enough to detect it as a problem?

wordsworthc: Python is as python does.. I think the more interesting part is that `pure-interface` didn't complain (which is _supposed_ to protect us from pythonic problems like this). Maybe we could consider a PR there to add async/sync checks on interface methods?

wordsworthc (approved):

l-macdonald (approved):

PR #71, opened 2025-06-17
add py.typed files.

mikecowie-seequent:
## Description

Mypy currently raises errors when invoked on a project using the common SDK package (and presumably the other packages).

Here py.typed files are added as specified in [PEP-561](https://peps.python.org/pep-0561/) so that type checkers like mypy will fully analyze the package when it's included in a project.

## Checklist


l-macdonald (approved):

wordsworthc (approved):

wordsworthc:
Don't forget to add the `package-matadata` entries

mikecowie-seequent:
@XavierMilesSeequent not just by adding the file. Adding the file just tells other projects that the SDK is meant to be type-hinted.

(mypy by default baulks over packages without either this file, or a `stubs` package), so when we have mypy enforcing we have to scatter this to make mypy pass. I think pyright is less fussy on this).

```python
from evo.aio import AioTransport # type: ignore[import-untyped]
from evo.common import APIConnector # type: ignore[import-untyped]
```

because (internal-repo) has mypy enforcing.

If you want my _opinion_, yeah I think it's a good idea for you to add an enforcing mypy or pyright check to make sure the type hints are consistent and accurate.

mikecowie-seequent:
>Apologies, I forgot this was changed to hatchling. I'm too used to setuptools data files support where package_data is used to be more explicit about which additional files should be included

All good! I suspected you might be thinking about some setuptools stuff, but also was wondering if it was something entirely different too :) .

PR #74, opened 2025-06-23
Add evo-sdk metapackage, which installs all other packages with dev dependencies

robbieaverill:
## Description

This PR renames the monorepo to `evo-sdk` and includes all optional dependencies (e.g. Jupyter notebooks).

The aim of this is to make it easy for people who want to get started with everything, including the notebooks, without forcing all users of the individual SDK packages to install the notebooks as well (for example). This will keep production environments more slim and free of optional dependencies while improving the early developer experience.

## Checklist

### DIFF
--- a/.github/workflows/publish-evo-sdk.yaml
+++ b/.github/workflows/publish-evo-sdk.yaml
@@ -29,7 +29,7 @@ jobs:
 
       - name: Build package
         shell: bash
-        run: uv build --package evo-sdk
+        run: uv build
 
       - name: Publish to PyPI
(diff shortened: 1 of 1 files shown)


packages/evo-sdk/pyproject.toml
wordsworthc, line 1: Why not just define `evo-sdk` in the root `pyproject.toml`? As a meta-package, there is no need for a `src`, and the readme, license, dependencies, etc. are already defined. All that should be needed is to rename `name = evo-python-sdk` -> `name = evo-sdk`

PR #76, opened 2025-06-25
Remove OIDC support and OIDC connector layer

robbieaverill:
## Description

WIP.

Aims:
* Remove OIDC entirely
* Replace with standard OAuth connector
* Refer to Bentley developer documentation for auth and apps rather than OAuth
* Ensure everything is working

Still to do:
* ~Replace `validate_*_token()` method with JWT library to validate claims: https://pyjwt.readthedocs.io/en/stable/usage.html#registered-claim-names~
* ~Refactor AccessToken model so it has access to the parsed token's claims as model properties (use an in-memory cache)~
* Ensure all notebooks are working
* Ensure all tests are working, and add new ones if necessary
* Tidy up, remove WIP
* Update documentation

## Checklist


packages/evo-sdk-common/src/evo/notebooks/authorizer.py
```diff
@@ -77,18 +73,10 @@ async def reuse_token(self) -> bool:
...
-                        client_id=self._connector.client_id,
-                        ignore_iat=True,
-                    )
-                except oauth.OAuthError as exc:
-                    logger.warning(f"Invalid token found in the environment file: {exc}")
-                    # The token is invalid. Clear it from the environment and return False.
-                    self._env.set_token(None)
-                    return False
+            #if isinstance(token, oauth.AccessToken):
+                # todo if token has expired return false
```
robbieaverill, line 77: Re-implement `token.validate_claims()` or whatever I renamed it to

packages/evo-sdk-common/src/evo/oauth/connector.py
```diff
@@ -23,72 +21,54 @@
...
 
 T = TypeVar("T", bound=AccessToken)
 
 
-class OIDCConnector:
-    """OpenID Connect client with endpoint discovery.
+ENDPOINTS = {
+    "token": "/connect/token",
+    "authorize": "/connect/authorize",
+}
```
robbieaverill, line 37:
```suggestion
```

packages/evo-sdk-common/src/evo/oauth/data.py
```diff
@@ -95,39 +84,32 @@ def __len__(self) -> int:
...
     """
 
     token_type: Literal["Bearer"]
+
     access_token: str
     """The access token issued by the authorization server."""
 
+    refresh_token: Optional[str] = None
+    """The refresh token, which can be used to obtain new access tokens using the same authorization grant. Refresh tokens are only returned if you request the "offline access" scope."""
+
```
robbieaverill, line 109: I'd like to add an init method to this model that actually parses and validates the JWT using https://pyjwt.readthedocs.io/en/stable/usage.html#registered-claim-names. It should use a memory cache property on the model, and we can connect the various properties like `issued_at` to the underlying JWT claims.

The actual response from IMS has `access_token`, `refresh_token` (if you have offline access requested), `expires_in` and `token_type`. Everything else would need to be accessed from the access token's claims.

packages/evo-sdk-common/src/evo/notebooks/authorizer.py
robbieaverill: Ok great, less work to do!

packages/evo-sdk-common/src/evo/oauth/data.py
robbieaverill: Following on from the comment above: https://github.com/SeequentEvo/evo-python-sdk/pull/76#discussion_r2167741814

Chris suggested we don't need to validate the JWT we receive at all, so in that case we can ignore this. SDK consumers can parse and extract data from the token if they want to.

packages/evo-sdk-common/src/evo/common/connector.py
```diff
@@ -235,6 +236,8 @@ async def call_api(
             response_type = default_response_type
 
         # Decode response.
+        logger.debug("RAW RESPONSE")
+        logger.debug(str(response.data))
```
wordsworthc, line 240: I also prefer not to split log messages that should be formatted together, because it is easy for them to get separated in the log output stream.

```suggestion
        logger.debug(f"HTTP response {response.status} ({response.reason}):\n{response.data}")
```

Food for thought:
* in the specific case of request + response, do we want to generate a correlation ID that we can include in debug messages?
* is it too much to include the raw response data in debug messages? Would it be sufficient to log the status code (and reason)?
* * `ApiConnector` is also used to transfer binary data, which can potentially be GB worth of bytes in the log output in this case

packages/evo-sdk-common/src/evo/notebooks/_consts.py
```diff
@@ -11,6 +11,7 @@
 
 DEFAULT_DISCOVERY_URL = "https://discover.api.seequent.com"
 DEFAULT_ISSUER_URL = "https://ims.bentley.com"
```
wordsworthc, line 13: Remove issuer URL

packages/evo-sdk-common/src/evo/oauth/data.py
```diff
@@ -95,39 +84,32 @@ def __len__(self) -> int:
...
     expires_in: Optional[int] = None
     """The lifetime in seconds of the access token."""
 
-    issued_at: datetime = Field(default_factory=_utcnow)
```
wordsworthc, line 125: Let's keep this. It defaults to a current UTC timestamp so that we can at least approximate an expiry time based on when the token was received.

PR #80, opened 2025-07-03
Adding list workspaces summary to the evo python sdk.

Gibson-Seequent:
## Description
Adding `list_workspaces_summary` method to `WorkspaceAPIClient`. This endpoint allows for optionally paginated calls for workspaces. Returning only id and name. To be used when list performance is important.

Supports all query parameters used in the standard `list_workspaces` method.

## Checklist


GriffinBaxterSeequent (approved):

sachinseizer (approved):
LGTM

PR #86, opened 2025-07-09
Add pre-commit hook to strip outputs from Jupyter notebooks

robbieaverill:
## Description

* Ensure the pre-commit hook for running the linter matches our Makefile commands
* Add pre-commit hook to remove all outputs from Jupyter notebooks before they can be committed

## Checklist


uv.lock
```diff
@@ -1,4 +1,5 @@
 version = 1
+revision = 2
```
robbieaverill, line 2: Changes are the result of a newer uv version

davidknight-seequent (approved):

PR #87, opened 2025-07-16
WS order_by allow both str and enum input

lokkochan:
## Description

order_by is taking in enum, but is not converted to string before processing them in next layer. This result in 422 as enum instead of str is being passed in. (Issue for python3.11 or higher)

Allow both str and enum input and client preprocess them into string if is enum being passed in

## Checklist

### DIFF
--- a/packages/evo-sdk-common/src/evo/workspaces/client.py
+++ b/packages/evo-sdk-common/src/evo/workspaces/client.py
@@ -173,7 +173,7 @@ class WorkspaceAPIClient:
         self,
         limit: int | None = None,
         offset: int | None = None,
-        order_by: dict[WorkspaceOrderByEnum, OrderByOperatorEnum] | None = None,
+        order_by: dict[WorkspaceOrderByEnum | str, OrderByOperatorEnum | str] | None = None,
         filter_created_by: UUID | None = None,
         created_at: str | None = None,
         updated_at: str | None = None,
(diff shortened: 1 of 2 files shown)


Gibson-Seequent (approved):

packages/evo-sdk-common/src/evo/workspaces/client.py
```diff
@@ -185,7 +185,14 @@ async def list_workspaces(
         if not offset:
             offset = 0
         if order_by:
-            parsed_order_by = ",".join([f"{op}:{field}" for field, op in order_by.items()])
+            parsed_order_by = ",".join(
```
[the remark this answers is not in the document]
robbieaverill: The Literal idea is a good one. Providing `{"name": "asc"}` is simple and easy, and Literal would give you the typing benefits still without having to play with all the enums and imports.

packages/evo-sdk-common/src/evo/workspaces/client.py
```diff
@@ -173,7 +174,9 @@ async def list_workspaces(
         self,
         limit: int | None = None,
         offset: int | None = None,
-        order_by: dict[WorkspaceOrderByEnum, OrderByOperatorEnum] | None = None,
+        order_by: dict[WorkspaceOrderByEnum, OrderByOperatorEnum]
+        | dict[Literal["name", "created_at", "updated_at", "user_role"], Literal["asc", "desc"]]
```
wordsworthc, line 178: DRY: I would be inclined to, at the very least, define these literal types at the top of the file so that any changes to supported fields or operators are contained to one place. That should also make it easier to read, IMO.

E.g.,
```python
WorkspaceOrderByLiteral: TypeAlias = Literal["name", "created_at", "updated_at", "user_role"]
OrderByOperatorLiteral: TypeAlias = Literal["asc", "desc"]
```

lokkochan: that looks way cleaner!!

robbieaverill (approved):
IMO, we can remove the enum typing and just have the new Literal types. Happy to merge in its current state though. Thanks @lokkochan!

PR #89, opened 2025-07-30
Object delete, restore, listing features, and timeouts

chriscunningham-seequent:
## Description

 - Added support for deleting and restoring objects
 - Added timeout options for individual endpoints
 - Added support for listing across an instance as a whole
 - Added support for `order_by` and filtering by `schema_id` in listing endpoints

## Checklist


packages/evo-objects/src/evo/objects/client.py
```diff
@@ -167,24 +190,39 @@ def _metadata_from_endpoint_model(self, model: GetObjectResponse | PostObjectRes
             version_id=model.version_id,
         )
 
-    async def list_objects(self, offset: int = 0, limit: int = 5000) -> Page[ObjectMetadata]:
+    async def list_objects(
+        self,
+        offset: int = 0,
+        limit: int = 5000,
+        order_by: dict[ObjectOrderByEnum | str, OrderByOperatorEnum] | None = None,
```
robbieaverill, line 197: In #87 we suggested that `str` could be defined as a `Literal` type rather than just `str`, to give it some more type safety. What do you think about implementing that approach here too?

packages/evo-sdk-common/src/evo/common/utils/data.py
```diff
@@ -0,0 +1,28 @@
...
+    if order_by is None:
+        return None
+    parts = []
+    for field, op in order_by.items():
+        if isinstance(field, Enum):
+            field = field.value
+        if isinstance(op, Enum):
+            op = op.value
+        parts.append(f"{op}:{field}")
+    return ",".join(parts)
```
robbieaverill, line 28: Good idea, would also be good to centralise this so it is used by #87 too if possible

robbieaverill (commented):
Looks like some really useful changes here, thank you @chriscunningham-seequent! I had a quick skim and I think there are some opportunities for consistency with recent changes to do with order parameters, which were merged in #87. Overall I don't have much else to provide for input, but happy to leave a green tick once you've had a chance to respond to these points.

packages/evo-objects/src/evo/objects/client.py
chriscunningham-seequent: For objects, you can order by any arbitrary property in the schema, so str is the only reasonable type there.

davidknight-seequent (approved):

l-macdonald (approved):

PR #91, opened 2025-08-11
fix: allow port number to be maintained to permit use of developmentstorage

mikecowie-seequent:
## Description

For internal development of evo services, we want to be able to run storage service quickly in test suite on a non-standard port.
Therefore the SDK should maintain the port in the URLs, if a port was provided as part of the base URL.

Have factored out the url parsing into its own util module to isolate the logic.

## Checklist


stephanwindischmann (approved):
Change looks good to me.

PR #92, opened 2025-08-20
Move audit out of all evo scope

lokkochan:
## Description

As audit scope is not configured for all entities that uses all_evo scope, moving the audit out will tackle the issue for those entity getting error during scope check.

## Checklist

### DIFF
--- a/packages/evo-sdk-common/pyproject.toml
+++ b/packages/evo-sdk-common/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "evo-sdk-common"
 description = "Python package that establishes a common framework for use by client libraries that interact with Seequent Evo APIs"
-version = "1.0.0"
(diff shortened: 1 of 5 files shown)


shivangj-10:
@lokkochan Thanks very much for the contribution. Can you please let us know the reason behind removing the evo.audit scope? It is a required scope to access audit service as well as a default scope provided for an Evo app in Common Application Registry (https://developer.bentley.com/).

lokkochan:
Thx for hving a look!! The major reason is that the addition of audit scope will cause some issues in clients that uses all_evo as they may are not configured to hv audit scope, eg, OAuth client. So removing audit from all_evo will address that. 
Can add it back later with a major update when we need it.

shivangj-10 (approved):
LGTM

PR #97, opened 2025-08-29
Extensible oauth scopes

wordsworthc:
## Description

The existing OAuth scopes implementation is very restrictive, and does not allow side-loading different and/or additional scopes with the provided OAuth authorizer implementations. This PR provides a scopes implementation that allows easily extending/customizing the authorizers in the SDK with different scopes, which can be useful in some cases.

The suggested change implements a `Scopes` type, as a subclass of `str`. `Scopes` overloads the `|` operator to combine multiple scopes into one `Scopes` instance, which is equivalent to a de-duplicated space-separated list of scopes with deterministic ordering. In this way it is functionally equivalent to combining and formatting the previous `OAuthScopes` members.

The `EvoScopes` is an `Enum` with the same members as `OAuthScopes`, which maintains the status quo in terms of a convenient way to access known Evo scopes. Thanks to this `Enum` also subclassing `Scopes`, `EvoScope` members can also be combined in the same manner as `OAuthScopes` allowed previously.

All method parameters that were previously annotated as `OAuthScopes` have been updated to `AnyScopes`, which is defined as `str | Scopes | EvoScopes`. It is worth noting that all three types satisfy `str`, but the inclusion of `Scopes` and `EvoScopes` is an explicit reference to the intended use.

Finally, `OAuthScopes` is provided as an alias to `EvoScopes` for backwards compatibility, but is marked as deprecated for type checking as the `EvoScopes` implementation should be preferred. I used the name `EvoScopes` in the new implementation because I think it better reflects the specificity of the scopes defined within.

## Checklist

### DIFF
--- a/packages/evo-sdk-common/pyproject.toml
+++ b/packages/evo-sdk-common/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "evo-sdk-common"
 description = "Python package that establishes a common framework for use by client libraries that interact with Seequent Evo APIs"
-version = "0.4.5"
+version = "0.5.0"
 requires-python = ">=3.10"
 license-files = ["LICENSE.md"]
 dynamic = ["readme"]


packages/evo-sdk-common/tests/oauth/test_client_credientials_authorizer.py
```diff
@@ -41,7 +41,7 @@ async def test_authorize(self):
         self.assert_token_equals(self.first_token)
         self.assert_fetched_token(
             grant_type="client_credentials",
-            scope="evo.discovery evo.workspace evo.blocksync evo.object evo.file",
+            scope="evo.blocksync evo.discovery evo.file evo.object evo.workspace",
```
wordsworthc, line 44: changed from sorting by Flag value (auto int) to sorting by string value

wordsworthc:
@davidknight-seequent done

l-macdonald (approved):

PR #99, opened 2025-09-03
Add support for configuring the close grace period for AioTransport

chriscunningham-seequent:
## Description

Add an optional parameter to AioTransport to allow configuration of the close grace period

## Checklist


julienpringotseequent (approved):
Should we keep the default to 250ms? I understand this is done for backward compatibility.

chriscunningham-seequent:
> Should we keep the default to 250ms? I understand this is done for backward compatibility.

In general clients will be using HTTPS, so they will need a grace period. 250 is the example value given in the aiohttp docs, though it helpfully says that the appropriate amount of time to wait "will vary from application to application".

PR #112, opened 2025-10-06
Make OAuthConnector retry token requests that respond with 502 or 504

jonny-norwine:
## Description

Fetching tokens from Bentley IMS can be flaky and respond with 502 and 504 errors. This introduces a retry policy to `OAuthConnector.fetch_token()` on 502 and 504 errors, following what @wordsworthc implemented in an internal package.

## Checklist


packages/evo-sdk-common/tests/oauth/test_connector.py
```diff
@@ -29,3 +75,84 @@ def test_endpoint(self, endpoint_type: Literal["authorize", "token"], expected:
...
+                    RESPONSE_500,
+                    RESPONSE_500,
+                    RESPONSE_500,
+                ),
+                1,
+                OAuthError,
+            ),
+        ]
+    )
+    async def test_fetch_token(
```
wordsworthc, line 135: Nice usage of `parameterized` & a `TestRequestHandler` too 😍 

One additional nice-to-have is the `@long_test` decorator, which will disable the tests unless the `CI` environment variable is set. It has been invaluable as a time-saver when running all tests locally. The GitHub action runners set the `CI` environment variable, so these tests will always be run there.

You can import the decorator like so:

```python
from evo.common.test_tools import long_test
```

Issue #123, opened 2025-10-19
Error running block model sample notebook


davidknight-seequent:
When I try to run the included sample notebook for `evo-blockmodels` I get this error:

```
Detail: 2 validation errors for 'HTTP Request'
body -> size_options
  Unable to extract tag using discriminator 'model_type' (type=union_tag_not_found; discriminator='model_type')
body -> coordinate_reference_system
  Value error, Invalid EPSG code. Value must either be a valid EPSG code in the form 'EPSG:nnnn', or be a valid WKT2 string. (type=value_error; error=Invalid EPSG code. Value must either be a valid EPSG code in the form 'EPSG:nnnn', or be a valid WKT2 string.)
```

PR #124, opened 2025-10-20
Fix errors when running the evo-blockmodels quickstart.ipynb notebook

BenLewis-Seequent:
## Description

Fixes https://github.com/SeequentEvo/evo-python-sdk/issues/123
Fix the sample notebook for the `evo-blockmodels` package, so it runs without any errors.

This includes the following fixes:
- explicitly pass the `model_type` property, so it is included in the request. `APIConnector` uses `exclude_unset=True`, previously the  `model_type` property wasn't being sent.
- use a CRS that block model service supports, i.e. one that is in meters or feet
- specify the column types of the `i`, `j`, and `k` columns as block model service expects the columns to be UInt32.
- fix examples trying to use the ID of the Geoscience Object, rather than the block model.

I also tweaked the example to produce a bit, to show off more behavior of the block model service.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Changelog
 
+## evo-sdk@v0.1.11
+### What's changed
+#### evo-sdk
+* Fix model_type parameter being excluded from the API requests and fix the sample notebook by @BenLewis-Seequent in https://github.com/SeequentEvo/evo-python-sdk/pull/124
+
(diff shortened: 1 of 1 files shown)


julienpringotseequent (approved):

GriffinBaxterSeequent (commented):
Could you please bump the versions for evo-blockmodels and the root evo-sdk, cheers :)

BenLewis-Seequent:
@GriffinBaxterSeequent I think I updated it correctly.

PR #136, opened 2025-10-31
Support downloading category data and attributes with DownloadedObject

BenLewis-Seequent:
## Description

Add methods to DownloadedObject to download data based on some common Geoscience Object Schema components, in particular category data and attributes. 

Examples of how these can be used
```python
# To download the "hole_id" column on downhole intervals
await downloaded_object.download_category_dataframe("hole_id")
```

```python
# To download and attribute named "Au_ppm", which correctly applies the nan_description values, and also automatically handles category attributes
await downloaded_object.download_attribute_dataframe("attributes[?name == 'Au_ppm']")
```

Note, for categories:
- the PyArrow array that is produced when using the `*_table` methods is a [DictionaryArray](https://arrow.apache.org/docs/python/generated/pyarrow.DictionaryArray.html), and 
- the pandas data type for the columns when using the `*_dataframe` methods is [CategoricalDtype](https://pandas.pydata.org/docs/user_guide/categorical.html). 

But the codes/indices used by the `DictionaryArray`/`CategoricalDtype` doesn't necessary match what is in the `values` table due to the category keys being able to be arbitrary int32/int64 and not just the index of the category.  

This also extends the other download methods:
 - to allow specifying which values to treat as nan, and
 - override the column names in the resultant table/dataframe. The column names in the parquet file aren't that useful and there is no standard for them.

Also, added MultiDownloadRequestHandler to help testing workflows to download multiple tables.

## Checklist


packages/evo-objects/src/evo/objects/client/object_client.py
```diff
@@ -182,63 +218,255 @@ def prepare_data_download(self, data_identifiers: Sequence[str | UUID]) -> Itera
...
+
+                arrays = []
+                for array in table.columns:
+                    if isinstance(array, pa.ChunkedArray):
+                        array = array.combine_chunks()
+                    null_scalar = pa.scalar(None, type=array.type)
+                    nan_value_array = pa.array(nan_values, type=array.type)
+                    arrays.append(pc.replace_with_mask(array, pc.is_in(array, nan_value_array), null_scalar))
+                return pa.Table.from_arrays(arrays, names=column_names)
+            else:
```
julienpringotseequent: else not needed.

julienpringotseequent (approved):

packages/evo-objects/src/evo/objects/client/object_client.py
```diff
@@ -182,63 +218,255 @@ def prepare_data_download(self, data_identifiers: Sequence[str | UUID]) -> Itera
...
+            t_size = category_info["table"]["length"] * 2  # Lookup tables always have 2 columns
+            split = _split_feedback(v_size, t_size)
+
+            values_table = await self.download_table(
+                category_info["values"],
+                nan_values=nan_values,
+                column_names=column_names,
+                fb=PartialFeedback(fb, start=0, end=split),
+            )
+            lookup_table = await self.download_table(category_info["table"], fb=PartialFeedback(fb, start=split, end=1))
```
wordsworthc: Probably not for this PR, but it would be nice to be able to download these in parallel. The main issue blocking that, I think, is that the current `PartialFeedback` implementation doesn't support concurrent operations.

I would be forever grateful if anyone was to come up with a solve for that - I expect it would significantly speed up many workflows I deal with.

wordsworthc (approved):

packages/evo-objects/src/evo/objects/client/object_client.py
BenLewis-Seequent: I avoided doing so precisely for that reason. There is going to be a lot of use cases where we will need to be able to update feedback concurrently, so that is an important problem to solve.

PR #137, opened 2025-11-03
Update all generated bindings in endpoints directories

GriffinBaxterSeequent:
## Description

These all now add headers with package names/versions. Endpoints directories have also been updated.

## Checklist

### DIFF
--- a/packages/evo-compute/pyproject.toml
+++ b/packages/evo-compute/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "evo-compute"
-version = "0.0.1rc1"
+version = "0.0.1rc2"
 requires-python = ">=3.10"
 license-files = ["LICENSE.md"]
 dynamic = ["readme"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "evo-sdk"
-version = "0.1.14"
+version = "0.1.15"
 description = "Python SDK for using Seequent Evo"
 requires-python = ">=3.10"
 dependencies = [


packages/evo-blockmodels/src/evo/blockmodels/endpoints/api/column_operations_api.py
```diff
@@ -22,15 +21,16 @@
...
 
 This code is generated from the OpenAPI specification for Block Model API.
-API version: 1.16.1
+API version: 1.41.3
 """
 
 from evo.common.connector import APIConnector
-from evo.common.data import EmptyResponse, RequestMethod  # noqa: F401
+from evo.common.data import RequestMethod
+from evo.common.utils import get_package_details
```
wordsworthc: It seems like the step that adds the `noqa` directive was missed. Which script did you use? IIRC, there is a powershell and a bash version, but the bash version, but the powershell version _should_ be the source of truth.

packages/evo-blockmodels/tests/test_create.py
```diff
@@ -275,7 +279,8 @@ async def test_create_block_model_with_data(self) -> None:
                     "Authorization": "Bearer <not-a-real-token>",
                     "Content-Type": "application/json",
                     "Accept": "application/json",
-                },
+                }
+                | header_metadata,
```
wordsworthc: `evo.common.test_tools.http.TestWithConnector` has `assert_request_made()` and `assert_any_request_made()` that wrap these `TestTransport` equivalents, inserting common headers into the expected headers so we don't have to update every test case for changes like this.

Perhaps you could update `TestWithConnector` for this use case?

packages/evo-blockmodels/src/evo/blockmodels/endpoints/models.py
```diff
@@ -90,67 +104,80 @@ class GeometryColumns(Enum):
...
-    decimal_char: StrictStr | None = Field(".", max_length=1, min_length=1, title="Decimal Char")
-    delimiter: StrictStr | None = Field(",", max_length=1, min_length=1, title="Delimiter")
-    file_format: Literal["csv"] = Field("csv", title="File Format")
-    quote_char: StrictStr | None = Field('"', max_length=1, min_length=1, title="Quote Char")
-    skip_rows: StrictInt | None = Field(0, ge=0, title="Skip Rows")
-    skip_rows_after_headers: StrictInt | None = Field(0, ge=0, title="Skip Rows After Headers")
+    column_name_mapping: Annotated[
+        list[ColumnNameMapping] | list[ColumnNameMappingGeneric] | None,
+        Field(title="Column Name Mapping"),
+    ] = []
```
wordsworthc: Oof.. mutable default value.. It is, however, equivalent to what it was before. Can you please raise a bug to look into how we can do this better?

packages/evo-blockmodels/src/evo/blockmodels/endpoints/models.py
```diff
@@ -510,13 +587,21 @@ class RotationAxis(Enum):
...
+    """
+    Type of sub-blocking
+    """
+
+    flexible = "flexible"
 
 
 class SizeOptionsFlexible(CustomBaseModel):
-    model_type: Literal["flexible"] = Field(title="Model Type")
+    model_type: Annotated[Literal["flexible"], Field(title="Model Type")] = "flexible"
```
wordsworthc: For `evo-blockmodels` in particular, these default values were recently removed due to an unintended interaction with the serialization logic. See #124 for more details

```suggestion
    model_type: Annotated[Literal["flexible"], Field(title="Model Type")]
```

packages/evo-blockmodels/src/evo/blockmodels/endpoints/models.py
```diff
@@ -553,8 +646,16 @@ class SizeOptionsFullySubBlocked(CustomBaseModel):
...
+    """
+    Type of sub-blocking
+    """
+
+    variable_octree = "variable-octree"
+
+
 class SizeOptionsOctree(CustomBaseModel):
-    model_type: Literal["variable-octree"] = Field(title="Model Type")
+    model_type: Annotated[Literal["variable-octree"], Field(title="Model Type")] = "variable-octree"
```
wordsworthc:
```suggestion
    model_type: Annotated[Literal["variable-octree"], Field(title="Model Type")]
```

packages/evo-files/src/evo/files/endpoints/models.py
```diff
@@ -63,7 +62,7 @@ class FileVersionResponse(CustomBaseModel):
     Metadata for specific file version
     """
 
-    created_at: Annotated[datetime, Field(title="Created At")]
+    created_at: Annotated[AwareDatetime, Field(title="Created At")]
```
wordsworthc: IIRC, the change to `AwareDatetime` has caused problems before. The fix was a slightly older version of the model generator, or there may even be a CLI parameter by now.

packages/evo-sdk-common/src/evo/common/utils/verison.py
```diff
@@ -36,4 +36,4 @@ def get_package_details(candidate: str) -> dict:
                 "version": package_metadata["version"],
             }
 
-    raise metadata.PackageNotFoundError(__package__)
+    return {}
```
wordsworthc: Is this change intentional? At first glance it seems invalid to return an empty dictionary (and perhaps a typed dict or named tuple with `name` and `version` would be more correct)

packages/evo-sdk-common/src/evo/workspaces/endpoints/api/discovery_api.py
```diff
@@ -54,7 +55,7 @@ class DiscoveryApi:
     def __init__(self, connector: APIConnector):
         self.connector = connector
 
-    async def v1_discovery_workspace_evo_identity_v1_discovery_get(
+    async def v1_discovery_evo_identity_v1_discovery_get(
```
wordsworthc: when generated clients have new method names like this, it would pay to check for references to the old name. Although in this case I don't think service discovery uses these generated clients (that code predates the schema)

packages/evo-sdk-common/src/evo/workspaces/endpoints/api/general_api.py
```diff
@@ -54,7 +55,7 @@ class GeneralApi:
     def __init__(self, connector: APIConnector):
         self.connector = connector
 
-    async def health_check_workspace_health_check_get(
+    async def health_check_health_check_get(
```
wordsworthc: Similarly, we don't actually rely on generated code for the health check endpoint

packages/evo-blockmodels/src/evo/blockmodels/endpoints/api/column_operations_api.py
GriffinBaxterSeequent: Looks like it should be adding them in both, not sure what went wrong.

packages/evo-blockmodels/src/evo/blockmodels/endpoints/models.py
GriffinBaxterSeequent: https://github.com/SeequentEvo/evo-python-sdk/issues/142 Done, cheers

packages/evo-files/src/evo/files/endpoints/models.py
GriffinBaxterSeequent: Thanks for this, yes this caused issues in other places so I've changed this back everywhere

packages/evo-sdk-common/src/evo/workspaces/endpoints/api/discovery_api.py
GriffinBaxterSeequent: Good point thanks, have reverted the name anyways as it doesn't need to change

wordsworthc: I think in this case it would be more beneficial to keep name changes like this, because it means the name in the schema has changed, and so the _next_ time someone runs the code generator we will encounter the same change

packages/evo-sdk-common/src/evo/common/utils/version.py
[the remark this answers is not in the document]
GriffinBaxterSeequent: Thank you, oops!

packages/evo-blockmodels/tests/test_create.py
GriffinBaxterSeequent: Unfortunately this doesn't seem to work as there's a few non-endpoints requests that are tested through these, and they don't have metadata headers in these.

GriffinBaxterSeequent: Unfortunately the best alternative would have to include passing through the header metadata to `assert_request_made` anyways as it needs the package the request came from.

packages/evo-sdk-common/src/evo/common/utils/version.py
wordsworthc: Or, perhaps: [version.patch](https://github.com/user-attachments/files/23486749/version.patch)

GriffinBaxterSeequent: Thanks, this is a lot better, only needed to modify slightly :)

wordsworthc (approved):

BenLewis-Seequent (approved):

rileyhowley-seequent (approved):
LGTM

PR #144, opened 2025-11-12
Blockmodel list models

rachelcarpenterseequent:
## Description
Start adding some more functionality to the blockmodel SDK.
This allows users to list all the blockmodels in their workspace, and allows the reading of all models, not just regular block models.

## Checklist

### DIFF
--- a/packages/evo-blockmodels/src/evo/blockmodels/client.py
+++ b/packages/evo-blockmodels/src/evo/blockmodels/client.py
@@ -13,7 +13,6 @@ from __future__ import annotations
 
 import asyncio
 import uuid
-from typing import List
 from uuid import UUID
 
 from evo import logging
@@ -167,7 +166,7 @@ class BlockModelAPIClient(BaseAPIClient):
 
             case _:
                 # Fall back to an informative error for unknown types
-                raise NotImplementedError(f"Size options type '{type(model.size_options).__name__}' is not supported")
+                raise NotImplementedError(f"Size options type '{type(model.size_options).__name__!r}' is not supported")
 
         return BlockModel(
             environment=self._environment,
@@ -267,11 +266,11 @@ class BlockModelAPIClient(BaseAPIClient):
         job_status = await self._poll_job_url(bm_id, job_id)
         return extract_payload(job_id, job_status, models.Version)
 
-    async def list_block_models(self) -> List[BlockModel]:
+    async def list_block_models(self) -> list[BlockModel]:
         """List block models in the current workspace.
 
-        Returns a list of `BlockModel` dataclasses for the workspace referenced by the client's
-        `Environment`. Limited to the first page of results.
+        Returns a list of `BlockModel`s for the workspace referenced by the client's
+        `Environment`. Limited to the first 100 results.
         """
         response = await self._metadata_api.list_block_models(
             workspace_id=str(self._environment.workspace_id),
@@ -280,7 +279,7 @@ class BlockModelAPIClient(BaseAPIClient):
 
         return [self._bm_from_model(m) for m in response.results]
 
-    async def list_all_block_models(self, page_limit: int = 100) -> List[BlockModel]:
+    async def list_all_block_models(self, page_limit: int | None = 100) -> list[BlockModel]:
         """Return all block models for the current workspace, following paginated responses.
 
         This method will page through the `list_block_models` endpoint using `offset` and `limit`
@@ -297,7 +296,7 @@ class BlockModelAPIClient(BaseAPIClient):
         page_limit = min(page_limit, 100)
 
         offset = 0
-        all_models: List[BlockModel] = []
+        all_models: list[BlockModel] = []
 
         while True:
             page = await self._metadata_api.list_block_models(
--- a/packages/evo-blockmodels/src/evo/blockmodels/data.py
+++ b/packages/evo-blockmodels/src/evo/blockmodels/data.py
@@ -67,19 +67,18 @@ class RegularGridDefinition(BaseGridDefinition):
 
 
 @dataclass(frozen=True, kw_only=True)
-class FullySubBlockedGridDefinition(BaseGridDefinition):
-    """A fully sub-blocked grid definition.
+class SubBlockedGridDefinition(BaseGridDefinition):
+    """A sub-blocked grid definition.
 
-    This represents a grid where each parent block is subdivided into a fixed number of regular
-    sub-blocks along each axis.
+    This represents a grid where each parent block is subdivided into a number of sub-blocks along each axis.
     """
 
     n_parent_blocks: list[int]
-    """Number of parent blocks along each axis."""
+    """Number of parent blocks along each axis. This must have 3 elements - `nx, ny, nz`."""
     n_subblocks_per_parent: list[int]
-    """Number of sub-blocks per parent block along each axis."""
+    """Number of sub-blocks per parent block along each axis. This must have 3 elements - `nx, ny, nz`."""
     parent_block_size: list[float]
-    """Size of the parent block along each axis."""
+    """Size of the parent block along each axis. This must have 3 elements - `x, y, z`."""
 
     def __post_init__(self):
         super().__post_init__()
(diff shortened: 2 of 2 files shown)


packages/evo-blockmodels/docs/examples/quickstart.ipynb
```diff
@@ -51,6 +51,33 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# list the first page\n",
+    "models = await service_client.list_block_models()\n",
+    "print(f\"Found {len(models)} models. First model id: {models[0].id}\")\n",
+    "\n",
+    "# fetch all models across pages with a page size of 50\n",
+    "all_models = await service_client.list_all_block_models(page_limit=5)\n",
```
tim-hamblin-seequent: Should the `page_limit` match the comment page size here?

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -219,6 +267,61 @@ async def _upload_data(self, bm_id: uuid.UUID, job_id: uuid.UUID, upload_url: st
...
+    async def list_all_block_models(self, page_limit: int = 100) -> List[BlockModel]:
+        """Return all block models for the current workspace, following paginated responses.
+
+        This method will page through the `list_block_models` endpoint using `offset` and `limit`
+        until all entries are retrieved. The `page_limit` is clamped to the service maximum (100).
+
+        :param page_limit: Maximum items to request per page (1..100). Defaults to 100.
+        :return: A list of `BlockModel` dataclasses for the workspace.
+        """
+        if page_limit is None:
```
tim-hamblin-seequent: Can `page_limit` be `None`? The typehint makes it seem like this isn't possible

BenLewis-Seequent (changes requested):
Looks good apart from inheritance structure of the grid definitions.

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -13,6 +13,7 @@
 
 import asyncio
 import uuid
+from typing import List
```
rohancrookbain-seequent: **issue:** `typing.List` is [deprecated](https://docs.python.org/3.13/library/typing.html#typing.List).
**suggestion:** could you please change all uses of `List` to the builtin `list`?

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -219,6 +267,61 @@ async def _upload_data(self, bm_id: uuid.UUID, job_id: uuid.UUID, upload_url: st
...
+            page = await self._metadata_api.list_block_models(
+                workspace_id=str(self._environment.workspace_id),
+                org_id=str(self._environment.org_id),
+                offset=offset,
+                limit=page_limit,
+            )
+
+            # Convert and append
+            for m in page.results:
+                all_models.append(self._bm_from_model(m))
```
rohancrookbain-seequent: **thought:** this is a perfectly good implementation, but it could be cool to also provide an async generator for these block models, so that users can access block models from the first page while later pages are still being downloaded.
I've very briefly looked into this and didn't get a simple way to do it though, so am happy for this to just be potential future work.

packages/evo-blockmodels/docs/examples/quickstart.ipynb
```diff
@@ -51,6 +51,33 @@
...
+    "print(f\"Total models returned: {len(all_models)}\")\n",
+    "\n",
+    "# inspect a model\n",
+    "bm = all_models[0]\n",
+    "print(bm.name)\n",
+    "print(bm.grid_definition.n_blocks)  # [nx, ny, nz]"
```
rohancrookbain-seequent: **nitpick:** it would be nice for these `print`s to have f-strings explaining what they're printing, similarly to the other ones in this block

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -118,8 +131,43 @@ def _bm_from_model(self, model: models.BlockModel | models.BlockModelAndJobURL)
...
+                    rotations=[(rotation.axis, rotation.angle) for rotation in model.block_rotation],
+                    n_parent_blocks=[n_parent_blocks.nx, n_parent_blocks.ny, n_parent_blocks.nz],
+                    n_subblocks_per_parent=[n_subblocks.nx, n_subblocks.ny, n_subblocks.nz],
+                    parent_block_size=[parent_block_size.x, parent_block_size.y, parent_block_size.z],
+                )
+
             case _:
-                raise NotImplementedError("Only regular models are supported at the moment")
+                # Fall back to an informative error for unknown types
+                raise NotImplementedError(f"Size options type '{type(model.size_options).__name__}' is not supported")
```
rohancrookbain-seequent: **nitpick:** it would be nice to use `!r` here, since that automatically deals with escaping special characters:
```suggestion
                raise NotImplementedError(f"Size options type {type(model.size_options).__name__!r} is not supported")
```
Of course it is very unlikely for special characters to be in a type name but this seems to me like good practice anyway.

packages/evo-blockmodels/src/evo/blockmodels/data.py
```diff
@@ -58,6 +66,48 @@ def __post_init__(self):
...
+    This represents a grid where each parent block is subdivided into a fixed number of regular
+    sub-blocks along each axis.
+    """
+
+    n_parent_blocks: list[int]
+    """Number of parent blocks along each axis."""
+    n_subblocks_per_parent: list[int]
+    """Number of sub-blocks per parent block along each axis."""
+    parent_block_size: list[float]
+    """Size of the parent block along each axis."""
```
rohancrookbain-seequent: **nitpick:** maybe it would be good to document here that these must have 3 values?

rohancrookbain-seequent (commented):
I agree with @BenLewis-Seequent's comment (which I feel is blocking); in addition I have a few comments (most if not all of which are non-blocking).

packages/evo-blockmodels/src/evo/blockmodels/client.py
rachelcarpenterseequent: I believe the file listing works in the same way as this, so I would prefer to leave that as a suggestions across the SDKs as a future improvement. Filed https://github.com/SeequentEvo/evo-python-sdk/issues/146 to capture this

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -219,6 +267,61 @@ async def _upload_data(self, bm_id: uuid.UUID, job_id: uuid.UUID, upload_url: st
...
+        `Environment`. Limited to the first page of results.
+        """
+        response = await self._metadata_api.list_block_models(
+            workspace_id=str(self._environment.workspace_id),
+            org_id=str(self._environment.org_id),
+        )
+
+        return [self._bm_from_model(m) for m in response.results]
+
+    async def list_all_block_models(self, page_limit: int = 100) -> List[BlockModel]:
```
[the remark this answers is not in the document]
wordsworthc: I disagree on this point - `page_limit: int | None` indicates that `None` is a valid value to pass here, whereas setting a default value makes it optional. IMO there is a lot of value in being more explicit about the default values in methods like this, and so I would lean toward optional values setting defaults of the most correct type, and explicitly not allowing `None` if it does not make sense.

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -118,8 +130,43 @@ def _bm_from_model(self, model: models.BlockModel | models.BlockModelAndJobURL)
...
+                    rotations=[(rotation.axis, rotation.angle) for rotation in model.block_rotation],
+                    n_parent_blocks=[n_parent_blocks.nx, n_parent_blocks.ny, n_parent_blocks.nz],
+                    n_subblocks_per_parent=[n_subblocks.nx, n_subblocks.ny, n_subblocks.nz],
+                    parent_block_size=[parent_block_size.x, parent_block_size.y, parent_block_size.z],
+                )
+
             case _:
-                raise NotImplementedError("Only regular models are supported at the moment")
+                # Fall back to an informative error for unknown types
+                raise NotImplementedError(f"Size options type '{type(model.size_options).__name__!r}' is not supported")
```
rohancrookbain-seequent: **nitpick:** — again, sorry! — the `!r` specifier adds quotes for you, so these are unnecessary:
```suggestion
                raise NotImplementedError(f"Size options type {type(model.size_options).__name__!r} is not supported")
```

wordsworthc (approved):

BenLewis-Seequent (approved):

PR #148, opened 2025-11-17
Blockmodel list models

rachelcarpenterseequent:
## Description

Start adding some more functionality to the blockmodel SDK.
This allows users to list all the blockmodels in their workspace, and allows the reading of all models, not just regular block models.

Note that this is just a repeat of https://github.com/SeequentEvo/evo-python-sdk/pull/144 which got closed when the base branch got deleted.

## Checklist


packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -118,8 +130,43 @@ def _bm_from_model(self, model: models.BlockModel | models.BlockModelAndJobURL)
...
+                    rotations=[(rotation.axis, rotation.angle) for rotation in model.block_rotation],
+                    n_parent_blocks=[n_parent_blocks.nx, n_parent_blocks.ny, n_parent_blocks.nz],
+                    n_subblocks_per_parent=[n_subblocks.nx, n_subblocks.ny, n_subblocks.nz],
+                    parent_block_size=[parent_block_size.x, parent_block_size.y, parent_block_size.z],
+                )
+
             case _:
-                raise NotImplementedError("Only regular models are supported at the moment")
+                # Fall back to an informative error for unknown types
+                raise NotImplementedError(f"Size options type '{type(model.size_options).__name__!r}' is not supported")
```
rohancrookbain-seequent: **nitpick:** the `!r` format specifier already includes quotes, so a second layer is unnecessary:
```suggestion
                raise NotImplementedError(f"Size options type {type(model.size_options).__name__!r} is not supported")
```

Issue #133, opened 2025-10-30
Block model SDK needs more features


davidknight-seequent:
The current block model SDK is a great first step, but it needs to be enhanced to allow for more user workflows.

Examples:
- the ability to update existing columns, eg. for renaming a column or for changing the units of a column.
- the ability to delete columns
- the ability to list all versions of a block model

The `update_block_model_columns` function includes parameters for updates and deletes, but it doesn't actually allow these operations to be done as individual operations.

The `query_block_model_as_table` function includes a `version_uuid` parameter for filtering purposes, but without the ability to list all versions of a block model, how is a user supposed to obtain this uuid?

rachelcarpenterseequent:
Hi @davidknight-seequent 
I think the last feature mentioned in the description is currently in review. There are examples for all the new methods added in https://github.com/SeequentEvo/evo-python-sdk/blob/main/packages/evo-blockmodels/docs/examples/quickstart.ipynb (check out the final PR for listing blockmodel versions if that hasn't landed yet).

Does that cover the functionality you were looking for? Are there improvements to these new methods that you can see?

PR #153, opened 2025-11-19
Add methods for listing blockmodel versions.

rachelcarpenterseequent:
## Description

This adds two methods for listing the versions of a blockmodel. One for listing just the first page of results, one for interating through all the pages.
Includes an example in the quickstart notebook.
Also cleaned up some of the other notebook cells so that they can be run out of order.

I believe this is the last requested feature on https://github.com/SeequentEvo/evo-python-sdk/issues/133

## Checklist

### DIFF
--- a/packages/evo-blockmodels/tests/test_list_block_models.py
+++ b/packages/evo-blockmodels/tests/test_list_block_models.py
@@ -199,8 +199,6 @@ class TestListBlockModels(TestWithConnector, TestWithStorage):
         self.assertEqual([v.version_id for v in result], [3, 2, 1])
 
     async def test_list_all_versions_empty_returns_empty(self) -> None:
-        import json
-
         bm_id = uuid.uuid4()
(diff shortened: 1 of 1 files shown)


packages/evo-blockmodels/tests/test_list_block_models.py
```diff
@@ -100,3 +120,92 @@ async def test_list_all_block_models_paginates_and_returns_all(self) -> None:
...
+            ),
+        ]
+        self.transport.request.side_effect = responses
+
+        result = await self.client.list_all_versions(bm_id, page_limit=2)
+        self.assertEqual(len(result), 3)
+        self.assertEqual([v.version_id for v in result], [3, 2, 1])
+
+    async def test_list_all_versions_empty_returns_empty(self) -> None:
+        import json
```
BenLewis-Seequent: Isn't json already imported in this module?

PR #154, opened 2025-11-19
Helpers to upload category data

BenLewis-Seequent:
## Description

Added methods `upload_category_table`/`upload_category_dataframe` to upload category tables and data frames. These methods extracts the indices and lookup table from the data, and uploads those to the Geoscience Object Service, returning the info that can be used within the geoscience object schemas.

This also adds support for specifying the table format or formats to allow when saving or uploading data. Which is often helpful, as you often know the table format you want to produced. 

## Checklist

### DIFF
--- a/packages/evo-sdk-common/src/evo/common/utils/feedback.py
+++ b/packages/evo-sdk-common/src/evo/common/utils/feedback.py
@@ -12,7 +12,7 @@
 import functools
 from collections.abc import Callable, Iterator, Sequence
 from threading import Lock
-from typing import Callable, TypeVar
+from typing import TypeVar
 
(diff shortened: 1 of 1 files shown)


ben-meneghini-seequent (approved):
Reviewed with @RyanMillerSeequent LGTM

wordsworthc:
FWIW, I would be _more_ in favour of a `ParquetSaver`/`ParquetUploader` implementation, as the intent with the `evo.objects.parquet` module was to begin deprecating loading (and saving eventually...) parquet data via `evo.objects.utils`. That is why `KnownTableFormat.load_table()` is marked as deprecated.

PR #163, opened 2025-12-02
Add IContext class to encapsulate all configurations required for Evo SDK clients

BenLewis-Seequent:
## Description

Add a interface `IContext`. and implementation `StaticContext` to `evo-sdk-common` to encapsulate all configuration classes need to the various Evo SDK client. That is it encapsulates:
- transport and authorizer classes
- optional cache
- the hub, org, and workspace to use (these are optional, but depending on the SDK client that it is being used with, missing values may raise a `ContextError` exception)

This avoids having to pass around multiple different objects to construct a client(e.g `ColormapAPIClient`).. Most of the clients now have a class method `from_context`, which constructs the client from a `IContext` object. The `ServiceManager`/`ServiceManagerWidget` now also implement the `IContext` interface. So `ColormapAPIClient.from_context(manager)` can be used to construct the `ColormapAPIClient` client from a notebook.

The clients that don't have a `from_context` are:
- `DiscoveryAPIClient`, as the API Connector it takes need to use the discovery base URL instead of the hub. Though I could imagine `EvoContext` could have a specific method `get_discovery_connect` to construct a API Connector this purpose which reuses the transport and authorizer.
- `JobClient`, as it is normally constructed by using `JobClient.submit` classmethod. We could add an overload for `JobClient.submit`that takes an `EvoContext`, but that would complicate that classmethod.

## Checklist

### DIFF
--- a/packages/evo-sdk-common/src/evo/common/interfaces.py
+++ b/packages/evo-sdk-common/src/evo/common/interfaces.py
@@ -11,10 +11,10 @@
 
 from __future__ import annotations
 
-import uuid
 from pathlib import Path
 from types import TracebackType
 from typing import TYPE_CHECKING
+from uuid import UUID
 
 from pure_interface import Interface
 
@@ -197,7 +197,7 @@ class IContext(Interface):
         :raises ContextError: If the context does not have sufficient information to create an Environment.
         """
 
-    def get_org_id(self) -> uuid.UUID:
+    def get_org_id(self) -> UUID:
         """Gets the organization ID associated with this context.
 
         :return: The organization ID.
--- a/packages/evo-sdk-common/src/evo/notebooks/widgets.py
+++ b/packages/evo-sdk-common/src/evo/notebooks/widgets.py
@@ -239,12 +239,13 @@ class ServiceManagerWidget(widgets.HBox, IContext, metaclass=_ServiceManagerWidg
         :param cache: The cache to use for storing tokens and other data.
         """
         self._authorizer = authorizer
+        self._cache = cache
         self._service_manager = ServiceManager(
             transport=transport,
             authorizer=authorizer,
             discovery_url=discovery_url,
+            cache=cache,
         )
-        self._cache = cache
         env = DotEnv(cache)
 
         self._btn = build_button_widget("Sign In")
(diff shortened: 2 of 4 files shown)


wordsworthc:
I think the main thing that jumps out at me here is the amount of crossover with the existing `evo.service_manager.ServiceManager`, and that the primary interfaces (from what I can tell) are nearly identical.

I acknowledge that the purposes and implementation are different enough to matter, i.e. `ServiceManager` is useful where the Evo hub URLs, org, workspace, are unknown or dynamic, whereas the `EvoContext` you propose seems to be for environments where these variables are known static values.

I think perhaps a better way to abstract this is to define an interface `IContext` that both variants conform to, so that either can be used as appropriate, without adapting between. From what I can see in this PR, I think the primary interface would be along the lines of the following?

```python
class IContext(Interface):
    def get_environment(self) -> Environment: ...

    def get_connector(self) -> APIConnector: ...

    def get_cache(self) -> ICache: ...
```

Then comes the where. I think an interface in `evo.common.interfaces` (exposed in `evo.common`) is in-keeping with existing patterns in `evo-sdk-common`. The `ServiceManager` implementation was moved to its own submodule largely due to its dependency on `evo.discovery`, but I think a static context is small enough to keep inside `evo.common`, as you have done here.

I'm thinking the eventual layout might look like the following:

> Add to `evo.common.interfaces`

```python
class IContext(Interface):
    def get_environment(self) -> Environment: ...

    def get_connector(self) -> APIConnector: ...

    def get_cache(self) -> ICache: ...
```

> Update `evo.common.context`

```python
# Renamed from `EvoContext` to reflect its status as a static variant of `IContext`
class StaticContext(IContext):
    ...  # `EvoContext` implementation from this PR
```

> Update `evo.service_manager.manager`

```python
class ServiceManager(IContext):
    ...  # Existing `ServiceManagerImplementation`
    
    # Also add `get_cache()` implementation
```

<details>
<summary>bonus round</summary>

> `evo.common.exceptions.SelectionError` becomes a sub-class of `evo.common.exceptions.ContextError`

```python
class SelectionError(ContextError):
    """Raised when a selection error occurs."""
```

> add the `from_context()` class method constructor to `evo.common.service.BaseAPIClient`, giving most current API clients the functionality "for free"

```python
class BaseAPIClient:
    ...  # Existing implementation

    @classmethod
    def from_context(cls, context: IContext) -> Self:
        ...  # Implementation from this PR
```

</details>

PR #164, opened 2025-12-03
Regular 3D grid typed object

BenLewis-Seequent:
## Description

This introduces "typed objects", using regular 3D grid as an example.

The existing interfaces in the `evo-objects` SDK is around handling Geoscience Objects generically with respect to the object type. Though in a lot of use cases, the user of the SDK knows which type of object they want to handle, whether that is a point set, regular 3D grid, etc. Providing classes that are specific to a particular type of object, provides better user experience as the user does not need to know the details about the schemas, and can just use auto complete or refer to the class directly. It also gives the ability to do more validation.

### Examples

Creating an regular 3D grid:

```python
grid = Regular3DGridData(
    name="Test Grid21",
    origin=Point3(0, 0, 0),
    size=Size3i(10, 10, 5),
    cell_size=Size3d(2.5, 5, 5),
    cell_data=pd.DataFrame(
        {
            "value": np.random.rand(10 * 10 * 5),
            "cat": pd.Categorical.from_codes(np.random.choice(range(4), size=10 * 10 * 5), ["a", "b", "c", "d"]),
        }
    ),
    vertex_data=pd.DataFrame(
        {
            "elevation": np.random.rand(11 * 11 * 6),
        }
    ),
    rotation=Rotation(90, 0, 0),
)
created_grid = await Regular3DGrid.create(manager.get_context(), grid)
```

Download:
```python
grid = await Regular3DGrid.from_reference(manager.get_context(), created_grid.metadata.url)
name = grid.name
cells_df = await grid.cells.as_dataframe()
```

Update:
```python
grid.name = "Updated Test Grid"
await grid.cells.set_dataframe(
    pd.DataFrame(
        {
            "value": np.ones(10 * 10 * 5),
        }
    )
)
# Update the object to create a new version with the modified properties
await grid.update()
```

### Schema versions

This is designed to gracefully handle multiple major versions of object schemas, through the use of adapters that map the details of the schemas to the class properties. This isn't fully fleshed out yet, but these details are not part of the public(stable) interface. 
This also only inspects schema properties upon access, meaning it should be able to handle newer minor versions of the object schemas, as long as the parts of the object that is accessed isn't using values added in the newer versions(i.e. new enum values).

## Checklist

### DIFF
--- a/packages/evo-objects/docs/examples/typed-objects.ipynb
+++ b/packages/evo-objects/docs/examples/typed-objects.ipynb
@@ -105,7 +105,7 @@
    "id": "8",
    "metadata": {},
    "source": [
-    "To download an existing 'regular-3d-grid' object, an inspect it, you can use the `Regular3DGrid.from_reference` class method. This method takes the context and the URL of the object to download.\n",
+    "To download an existing 'regular-3d-grid' object, and inspect it, you can use the `Regular3DGrid.from_reference` class method. This method takes the context and the URL of the object to download.\n",
     "\n",
     "For the example, we will download the object we just created."
    ]
@@ -160,7 +160,7 @@
    "id": "13",
    "metadata": {},
    "source": [
-    "# Updating a regular 3D grid object"
+    "## Updating a regular 3D grid object"
    ]
   },
   {
@@ -170,7 +170,7 @@
    "source": [
     "To update an existing 'regular-3d-grid' object, we can modify the properties of the `Regular3DGrid` instance, and then call the `update` method to submit the changes to the Geoscience Object service. This will create a new version of the object with the updated properties.\n",
     "\n",
-    "Updating the cell or vertex data can be done by calling the `set_dataframe` method on the `cells` or `vertices` attributes respectively, this will upload the new data to the Geoscience Object service, but the object will remain the same version until the `update` method is called."
+    "Updating the cell or vertex data can be done by calling the `set_dataframe` method on the `cells` or `vertices` attributes respectively. This will upload the new data to the Geoscience Object service, but the object will remain the same version until the `update` method is called."
    ]
   },
   {


packages/evo-objects/src/evo/objects/typed/_property.py
```diff
@@ -0,0 +1,82 @@
...
+    def __get__(self, instance: WithDocument | None, owner: type[WithDocument]) -> Any:
+        if instance is None:
+            return self
+
+        value = instance.search(self._jmespath_expr)
+        if value is None and self._default_factory is not None:
+            return self._default_factory()
+        if isinstance(value, (jmespath.JMESPathArrayProxy, jmespath.JMESPathObjectProxy)):
+            value = value.raw
+        return self._typed_adapter.validate_python(value)
```
wordsworthc: If we rely on pydantic to handle default values, this becomes:
```suggestion
        if isinstance(value, (jmespath.JMESPathArrayProxy, jmespath.JMESPathObjectProxy)):
            value = value.raw
        return self._typed_adapter.validate_python(value)
```

RyanMillerSeequent (approved):
LGTM

rileyhowley-seequent (approved):
This is great work @BenLewis-Seequent and will be very beneficial for users. I just had a couple of questions but nothing blocking.

packages/evo-objects/src/evo/objects/typed/_utils.py
```diff
@@ -0,0 +1,195 @@
...
+
+    return _response_to_downloaded_object(response, environment, connector, context.get_cache())
+
+
+async def download_geoscience_object(context: IContext, reference: ObjectReference) -> DownloadedObject:
+    return await DownloadedObject.from_reference(
+        connector=context.get_connector(),
+        reference=reference,
+        cache=context.get_cache(),
+    )
```
wordsworthc: Can we add a `from_context()` method to `DownloadedObject`?

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -0,0 +1,555 @@
...
+    If None, this class is considered abstract and cannot be instantiated directly.
+    """
+
+    creation_schema_version: ClassVar[SchemaVersion | None] = None
+    """The version of the Geoscience Object schema to use when creating new objects of this type.
+
+    If None, this class can't create a new Geoscience Object, but can still load an existing one.
+    """
+
+    def __init__(self, context: IContext, obj: DownloadedObject) -> None:
```
wordsworthc: This seems weird to me because everything that `IContext` provides is available in `DownloadedObject`. After digging in it seems `IContext` is needed for uploading new parquet data? Could we accept accessing the internals of `DownloadedObject` (`DownloadedObject._connector` and `DownloadedObject._cache`) for this?

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -0,0 +1,555 @@
...
+class BaseObject(_BaseObject):
+    """Base object for all Geoscience Objects, containing common properties."""
+
+    name: str = SchemaProperty("name", TypeAdapter(str))
+    description: str | None = SchemaProperty("description", TypeAdapter(str | None))
+    tags: dict[str, str] = SchemaProperty("tags", TypeAdapter(dict[str, str]), default_factory=dict)
+    extensions: dict = SchemaProperty("extensions", TypeAdapter(dict), default_factory=dict)
+
+    @classmethod
+    def create(
```
wordsworthc:
```suggestion
    async def create(
```

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -0,0 +1,555 @@
...
+
+        The type of Geoscience Object created is determined by the type of `data` provided.
+
+        :param context: The context containing the environment, connector, and cache to use.
+        :param data: The data that will be used to create the object.
+        :param parent: Optional parent path for the object.
+        :param path: Full path to the object, can't be used with parent.
+        """
+        object_type = cls._get_object_type_from_data(data)
+        return object_type._create(context, data, parent, path)
```
wordsworthc:
```suggestion
        return await object_type._create(context, data, parent, path)
```

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -0,0 +1,555 @@
...
+
+        :raises ValueError: If the bounding box cannot be computed from the datasets.
+        """
+        raise NotImplementedError("Subclasses must implement compute_bounding_box to derive bounding box from data.")
+
+
+class BaseSpatialObject(BaseObject):
+    """Base class for all Geoscience Objects with spatial data."""
+
+    _bbox_typed_adapter: ClassVar[TypeAdapter[BoundingBox]] = TypeAdapter(BoundingBox)
```
wordsworthc: nitpick:
```suggestion
    _bbox_type_adapter: ClassVar[TypeAdapter[BoundingBox]] = TypeAdapter(BoundingBox)
```

packages/evo-objects/src/evo/objects/typed/dataset.py
```diff
@@ -0,0 +1,496 @@
...
+            assign_jmespath_value(self._document, adapter.values_path, table_info)
+        elif isinstance(adapter, CategoryTableAdapter):
+            category_info = await data_client.upload_category_dataframe(
+                df[adapter.column_names],
+            )
+            assign_jmespath_value(self._document, adapter.category_data_path, category_info)
+        else:
+            raise ValueError(f"Unsupported adapter type: {type(adapter)}")
+
+    async def set_dataframe(self, df: pd.DataFrame, fb: IFeedback = NoFeedback) -> None:
```
wordsworthc: One line of discussion in the data converters initiative has been around separating serializing parquet data (to disk, perhaps) from uploading it to the API.

packages/evo-objects/src/evo/objects/typed/dataset.py
```diff
@@ -0,0 +1,496 @@
...
+    ) -> Self:
+        # Create an empty attribute list
+        assign_jmespath_value(document, dataset_adapter.attributes_adapter.attribute_list_path, [])
+
+        dataset = cls(document, dataset_adapter, context)
+        if data is not None:
+            await dataset.set_dataframe(data)
+        return dataset
+
+    def validate(self) -> None:
```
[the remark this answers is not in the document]
BenLewis-Seequent: The problem with validation on assignment is that the properties aren't necessarily orthogonal to each other. So navigating from one valid state to another might not be easy to do if all intermediate states also need to be valid.  For example in a regular grid where the number of rows in the attributes needs to match the number of blocks in the definition, all attributes must be cleared first, if we wanted to change the size of the grid.

This would likely mean defining methods that can updates all of the related properties at once, to avoid this complexity, i.e. the grid size and attributes. I'm unsure how well that would scale with more complex validation.

packages/evo-objects/src/evo/objects/typed/_property.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class SchemaProperty(Generic[_T]):
+    """Descriptor for data within a Geoscience Object schema.
+
+    This can be used on either typed objects classes or dataset classes.
+    """
+
+    def __init__(
+        self, jmespath_expr: str, typed_adapter: TypeAdapter[_T], default_factory: Callable[[], _T] | None = None
```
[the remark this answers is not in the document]
BenLewis-Seequent: Done, changed to automatically use the annotation on the fields instead, so now the pydantic Field class is now passed to the TypeAdapter.

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -0,0 +1,555 @@
...
+        :param reference: The reference of the object to create or replace.
+        :param data: The data that will be used to create the object.
+        """
+        object_type = cls._get_object_type_from_data(data)
+        return await object_type._replace(context, reference, data, create_if_missing=True)
+
+
+class ConstructableObject(BaseObject, Generic[_T]):
+    # The class methods in this class technically violate Liskov Substitution Principle,
+    # as they narrow the type of the 'data' parameter.
```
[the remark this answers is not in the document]
BenLewis-Seequent: Because I didn't want to make BaseObject generic, as that would complicate other using the class else where. I have removed this at least for now to simplify things, but does mean the type of the 'data' parameters isn't as precise as desired.

packages/evo-objects/src/evo/objects/typed/dataset.py
BenLewis-Seequent: From what I have been involved with, the separation discussions was more about separating the construction of the data objects from saving/uploading to the API. Which the data objects like `Regular3DGridData` is meant to solve.

Though I wonder whether we should have an API to save the data to the cache then upload any un-uploaded data when we process the update. Though for most users they shouldn't need to care about the details. The advantage of just saving the data and leaving the uploading till later is the save method wouldn't need to be async.

packages/evo-objects/docs/examples/typed-objects.ipynb
```diff
@@ -0,0 +1,222 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "14",
+   "metadata": {},
+   "source": [
+    "To update an existing 'regular-3d-grid' object, we can modify the properties of the `Regular3DGrid` instance, and then call the `update` method to submit the changes to the Geoscience Object service. This will create a new version of the object with the updated properties.\n",
+    "\n",
+    "Updating the cell or vertex data can be done by calling the `set_dataframe` method on the `cells` or `vertices` attributes respectively, this will upload the new data to the Geoscience Object service, but the object will remain the same version until the `update` method is called."
```
l-macdonald:
```suggestion
    "Updating the cell or vertex data can be done by calling the `set_dataframe` method on the `cells` or `vertices` attributes respectively. This will upload the new data to the Geoscience Object service, but the object will remain the same version until the `update` method is called."
```

PR #173, opened 2025-12-18
Add fallback to unknown and add test

AdamFreemanSeequent:
## Description

Allow for a fallback value in SDK usage metrics if package metadata is not available.

## Checklist

### DIFF
--- a/packages/evo-sdk-common/src/evo/common/utils/version.py
+++ b/packages/evo-sdk-common/src/evo/common/utils/version.py
@@ -50,7 +50,7 @@ def get_package_details(candidate: str) -> PackageDetails:
             candidate, *_ = candidate.rpartition(".")
         else:
             return PackageDetails(
-                name=package_metadata.get("name", "evo-sdk-common"),
+                name=package_metadata.get("name", "evo-sdk"),
                 version=package_metadata.get("version", "0.0.0"),
             )
 
--- a/packages/evo-sdk-common/tests/common/utils/test_version.py
+++ b/packages/evo-sdk-common/tests/common/utils/test_version.py
@@ -62,7 +62,7 @@ class TestGetPackageDetails(unittest.TestCase):
         with patch("evo.common.utils.version.metadata.metadata", return_value=mock_metadata):
             result = get_package_details("missing.both")
 
-        self.assertEqual(result.name, "evo-sdk-common")
+        self.assertEqual(result.name, "evo-sdk")
         self.assertEqual(result.version, "0.0.0")
 
 


packages/evo-sdk-common/src/evo/common/utils/version.py
```diff
@@ -50,8 +50,8 @@ def get_package_details(candidate: str) -> PackageDetails:
             candidate, *_ = candidate.rpartition(".")
         else:
             return PackageDetails(
-                name=package_metadata["name"],
-                version=package_metadata["version"],
+                name=package_metadata["name"] if "name" in package_metadata else "unknown",
+                version=package_metadata["version"] if "version" in package_metadata else "unknown",
```
wordsworthc: `Mapping.get(...)` is probably cleaner, IMO

```suggestion
                name=package_metadata.get("name", "unknown"),
                version=package_metadata.get("version", "unknown"),
```

packages/evo-sdk-common/tests/common/utils/test_version.py
```diff
@@ -0,0 +1,77 @@
...
+        """Test that get_package_details falls back to evo-sdk-common when no package is found."""
+
+        def mock_metadata_side_effect(candidate: str) -> MagicMock:
+            if candidate == "evo-sdk-common":
+                mock_meta = MagicMock()
+                mock_meta.__getitem__.side_effect = lambda key: {
+                    "name": "evo-sdk-common",
+                    "version": "0.1.0",
+                }[key]
+                mock_meta.__contains__.side_effect = lambda key: key in {"name", "version"}
```
wordsworthc:
```suggestion
                mock_meta = {
                    "name": "evo-sdk-common",
                    "version": "0.1.0",
                }
```

packages/evo-sdk-common/src/evo/common/utils/version.py
wordsworthc: Also, because this is provided by `evo-sdk-common`, do we want to default to the package name `evo-sdk-common`? (similar to the other fallback)

l-macdonald (approved):

packages/evo-sdk-common/src/evo/common/utils/version.py
AdamFreemanSeequent: Ah yes, @GriffinBaxterSeequent pointed that out as well. I have updated this including using the `evo-sdk-common` and version of `0.0.0` as suggested by Griffin also.

wordsworthc (approved):

packages/evo-sdk-common/src/evo/common/utils/version.py
```diff
@@ -50,8 +50,8 @@ def get_package_details(candidate: str) -> PackageDetails:
             candidate, *_ = candidate.rpartition(".")
         else:
             return PackageDetails(
-                name=package_metadata["name"],
-                version=package_metadata["version"],
+                name=package_metadata.get("name", "evo-sdk-common"),
```
GriffinBaxterSeequent: Sorry to hold this up, but I'm just not sure about using `evo-sdk-common` here when we don't even know the version. I'd personally rather `evo-sdk` referencing the monorepo package name, but even then I think `unknown` is the way to go. All in all, happy with what the majority thinks.

l-macdonald: thats a good point. Maybe evo-sdk would be better?

AdamFreemanSeequent: Sounds good to me.

PR #178, opened 2026-01-04
Escape error message before _build_redirect_html_failed

GriffinBaxterSeequent:
## Description

Escaping this error message before _build_redirect_html_failed is called so it can be detected as fixed by the code scanning.

## Checklist


AdamFreemanSeequent (approved):
Nice one @GriffinBaxterSeequent - sorry I sent you in the wrong direction

PR #198, opened 2026-02-03
Compute SDK PDAC Demo

denis-simo:
This is a non-insignificant set of changes so for the purpose of the review here is a summary.
The goal is to support geologist in their quest to do krigging via Seequent Evo.
The pull request brings quality of life changes for the existing typed objects:
* Pretty printing
* Links to Evo (even though many of them are not officially supported, they are what they are)
* Simplified access to the attributes and abstraction of schema complexity (`.locations.attributes`, `.vertices.attributes`, ... )

There is now support for all objects required as part of running compute tasks.
* Regular Block Models + Block model references as proxies - I would like to do Octree as part of a separate pull request, this one is already chunky. The user does not need to know about block model service,  block model reference acts as a proxy.
* Variograms / Ellipsoids - this object includes basic way to create a mesh for visualisation in a notebook.
* BlockSync Reports creation

Further on we believe and will show are workflow at PDAC where Copilot will be used to generate notebooks to help with krigging workflow, as such a lot of instruction files were added to support the user development with a simple prompt. They are not perfect, they are not optimal, they will change as more workflows are added and extended - but they work pretty well and they are a good starting point.

There are a few things outstanding:
- [ ] Need to do the review of Stylesheet. Seequent style vs style that works a bit better in the notebooks (current).
- [ ] Need to bump the packages.

Maybe to get maintainers happy we can consider to break this down into - will speed things up with getting things approved probably but increase the number of package version bumps:
1. HTML styles and pretty printing ( evo.common.styles.html module, _repr_html_() methods on BaseObject, Attributes, BlockModelAttributes, TaskResult, TaskResults, Variogram )
2. Block model typed object (Regular)
3. Ellipsoid + Variogram typed objects ( Ellipsoid, Variogram.get_ellipsoid(), get_principal_directions() )
4. Task Registry + Kriging ( Task registry, Attribute compute methods, Source/Target, SearchNeighborhood, KrigingParameters, KrigingResult )  + Notebook for running multiple
5. Documentation for auto-generation of notebooks via LLMs

### DIFF
--- a/packages/evo-colormaps/pyproject.toml
+++ b/packages/evo-colormaps/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "evo-colormaps"
 description = "Python SDK for using the Seequent Evo Colormap API"
-version = "0.0.2"
+version = "0.0.3"
 requires-python = ">=3.10"
 license-files = ["LICENSE.md"]
 dynamic = ["readme"]
--- a/packages/evo-colormaps/src/evo/colormaps/data.py
+++ b/packages/evo-colormaps/src/evo/colormaps/data.py
@@ -29,7 +29,7 @@ from evo.common import ResourceMetadata, ServiceUser
 
 def _validate_colors(colors: list[list[int]], minimum: int = 2, maximum: int = 1024):
     if not minimum <= len(colors) <= maximum:
-        raise ValueError("Colors must have between 2 and 1024 entries")
+        raise ValueError(f"Colors must have between {minimum} and {maximum} entries")
 
     for color in colors:
         if len(color) != 3 or any(not isinstance(value, int) or value < 0 or value > 255 for value in color):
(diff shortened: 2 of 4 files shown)


packages/evo-objects/src/evo/objects/typed/pointset.py
```diff
@@ -116,3 +117,23 @@ class PointSet(BaseSpatialObject):
...
+        """
+        return await self.locations._table.get_dataframe(fb=fb)
+
+    async def to_dataframe(self, *keys: str, fb: IFeedback = NoFeedback) -> pd.DataFrame:
+        """Get the full dataframe for the pointset, including coordinates and attributes.
+
+        Returns:
+            A DataFrame with 'x', 'y', 'z' columns and any additional attribute columns.
+        """
+        return await self.locations.to_dataframe(*keys, fb=fb)
```
denis-simo: @BenLewis-Seequent  wanted to double check with you which objects are fully in-in and which one are work in progress. 
However this sort of creature of life comforts are needed on grids as well and other objects.

PR #202, opened 2026-02-04
Typed objects qol

denis-simo:
The first set of changes to Evo typed objects based on user interviews.

* Easier access to data that is closer to numpy and pandas via `to_dataframe` (which effectively is an alias for `get_dataframe`)
* Easier access to attributes pointset.attributes['Au ppm'] instead of pointset.attributes['locations.Au ppm']  or pointset.locations.attributes['Au ppm']  
* If the user has forgotten to add `.json` let them be and add it for them.
* New accessors to create type objects that don't require the user to know what object type they want to create. 
Instead of `pointset = await PointSet.from_reference(manager, reference)` you cand do `poinset = await object_from_reference(manager, reference)`
* No expectation for the user to specify CRS, it defaults to something sensible if the user does not care about it.
* Basic links to Evo stuff. Do we officially support all of them - no, however users don't care and this package can be updated when we next change frontends or decide to support long term links in Evo (and not just blocksync block models)


packages/evo-sdk-common/src/evo/common/styles/html.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+    html_parts = [
+        STYLESHEET,
+        '<div class="evo">',
+        build_title(title),
+        build_table(rows),
+        extra_content,
+        '</div>'
+    ]
+    return ''.join(html_parts)
```
denis-simo: Image

Image

packages/evo-objects/src/evo/objects/typed/_data.py
```diff
@@ -33,6 +33,16 @@ async def get_dataframe(self, fb: IFeedback = NoFeedback) -> pd.DataFrame:
             raise DataLoaderError("Data was modified since the object was downloaded")
         return await self._obj.download_dataframe(self.as_dict(), fb=fb, column_names=self.data_columns)
 
+    async def to_dataframe(self, fb: IFeedback = NoFeedback) -> pd.DataFrame:
```
grantroch: I think we need to consider this change. We shouldn't have both of these methods around. Decide on one and move on. It seems like Ben already had conversations around this naming and perhaps they already considered `to` vs `get`.

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -283,21 +476,349 @@ def metadata(self) -> ObjectMetadata:
         """
         return self._obj.metadata
 
+
+    @property
+    def viewer_url(self) -> str:
```
grantroch: Can these be replaced by the shared functions in evo/common/urls?

grantroch (commented):
I don't understand the Annotation/Schema changes but looked at everything else.

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -283,21 +476,349 @@ def metadata(self) -> ObjectMetadata:
...
+
+        :param context: The context containing the environment, connector, and cache to use.
+        :param reference: The reference of the object to create or replace.
+        :param data: The data that will be used to create the object.
+        """
+        object_type = cls._get_object_type_from_data(data)
+        return await object_type._replace(context, reference, data, create_if_missing=True)
+
+
+class ConstructableObject(BaseObject, Generic[_T]):
```
BenLewis-Seequent: This doesn't appear to be used. 

I initially used this class so that the `data` parameters to the create/replace/create_or_replace methods had the correct and strict type. i.e. for `PointSet`, the 'data' parameters would be type `PointSetData`, so passing a `Regular3DGridData`, could be picked up by a type checker. I subsequently decided to remove it, for simplification.

packages/evo-objects/src/evo/objects/typed/_utils.py
```diff
@@ -191,3 +191,10 @@ async def replace_geoscience_object(
         response = await _update_geoscience_object(environment, objects_api, get_response.object.uuid, object_dict)
 
     return _response_to_downloaded_object(response, environment, connector, context.get_cache())
+
+async def download_geoscience_object(context: IContext, reference: ObjectReference) -> DownloadedObject:
+    return await DownloadedObject.from_reference(
+        connector=context.get_connector(),
+        reference=reference,
+        cache=context.get_cache(),
+    )
```
wordsworthc: This looks identical to `DownloadedObject.from_context()`

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -35,28 +42,172 @@
...
+        from evo.objects.typed import object_from_path
+
+        # Download latest version by path
+        obj = await object_from_path(context, "my-folder/pointset.json")
+        
+        # Download specific version
+        obj = await object_from_path(context, "my-folder/pointset.json", version="abc123")
+    """
+    version = "?version=" + version if version else ""
+    reference = ObjectReference(_get_url_prefix(context.get_environment()) + f"/path/{path}{version}")
```
wordsworthc:
```suggestion
    reference = ObjectReference.new(context.get_environment(), object_path=path, version_id=version)
```

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -35,28 +42,172 @@
...
+        raise ValueError(
+            f"No typed class found for sub-classification '{obj.metadata.schema_id.sub_classification}'. "
+            f"Available types: {list(_BaseObject._sub_classification_lookup.keys())}"
+        )
+    
+    return selected_cls(reference_context, obj)
 
-    _data_class_lookup: ClassVar[weakref.WeakValueDictionary[type[BaseObjectData], type[_BaseObject]]] = (
+
+async def object_from_path(
```
wordsworthc: The `object_from**` pattern seems somewhat repetitive. Can we deduplicate and perhaps settle on a `_BaseObject.from_context(..)` pattern that accepts:
* an `IContext`; and
* one of:
  * an `ObjectReference | str`, which may or may not include the version; or
  * an object `UUID`; or
  * an object path (`str`)
* and optionally if an object ID or path were used, a version ID.

Basically following the same pattern as `ObjectReference.new()` / deferring to its implementation for constructing a reference from UUID/path and version ID

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -283,21 +408,246 @@ def metadata(self) -> ObjectMetadata:
...
+
+        :param context: The context containing the environment, connector, and cache to use.
+        :param reference: The reference of the object to create or replace.
+        :param data: The data that will be used to create the object.
+        """
+        object_type = cls._get_object_type_from_data(data)
+        return await object_type._replace(context, reference, data, create_if_missing=True)
+
+
+
```
BenLewis-Seequent: Extra lines

packages/evo-objects/src/evo/objects/typed/base.py
```diff
@@ -72,31 +191,20 @@ class _BaseObject(SchemaModel):
     If None, this class can't create a new Geoscience Object, but can still load an existing one.
     """
 
-    def __init__(self, obj: DownloadedObject) -> None:
+    def __init__(self, context: IContext, obj: DownloadedObject) -> None:
```
wordsworthc: `IContext` was originally included in #164 , but removed in favour of using internals from `DownloadedObject` (see [this comment](https://github.com/SeequentEvo/evo-python-sdk/pull/164#discussion_r2625176194)). Seeing as all the prerequisite parts are definitely available, should we just make an interface adapter for `DownloadedObject` -> `IContext`? Or even implement it explicitly?

packages/evo-sdk-common/src/evo/common/styles/html.py
wordsworthc: I think I've made my position on core SDK vs utilities for notebook environments (and similar) clear elsewhere.

This file in particular is probably the biggest reason I think it's time to split it into a separate package that is installed by default with `evo-sdk`, but not with `evo-sdk-common`.

packages/evo-objects/src/evo/objects/typed/base.py
grantroch: It is repetitive but also explicit so the user knows the various ways they can create objects. As above, these functions avoid having to know the object type being loaded.

packages/evo-objects/src/evo/objects/typed/_data.py
grantroch: I also agree with Chris on the `from_dataframe` convention in these use cases. I think the old name was less confusing.

packages/evo-objects/src/evo/objects/typed/attributes.py
```diff
@@ -274,3 +295,31 @@ def validate_lengths(self, expected_length: int) -> None:
...
+            attribute_type = info["attribute_type"]
+            if attribute_type != "category":
+                attribute_str = f"{info['attribute_type']} ({info['values']['data_type']})"
+            else:
+                attribute_str = attribute_type
+            rows.append([info["name"], attribute_str])
+
+        # Use nested table for a clean header/row structure
+        table_html = build_nested_table(headers, rows)
+        return f'{STYLESHEET}<div class="evo">{table_html}</div>'
```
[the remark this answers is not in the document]
wordsworthc: An optional `_repr_html_` that depends on the additional module being available is certainly one way to do it, and we could defer as much of the logic as possible to the additional module. Although I did come across IPython docs for [formatters for third party types](https://ipython.readthedocs.io/en/latest/config/integrating.html#formatters-for-third-party-types), which could be an even cleaner way to separate them IMO

PR #208, opened 2026-02-16
Auto-generated SDK Documentation

GriffinBaxterSeequent:
## Description

Adds auto-generated SDK documentation using MkDocs and mkdocstrings-python. HTML files are output containing documentation of all clients, which can be edited via `mkdocs/api_clients.txt`. Many customisations have been made to suit the format desired for the Developer Portal, including a custom directory structure, parsing the sphinx docstring style, using the member order of client files, and more.

The docs can be updated by running `mkdocs build` in the `mkdocs/` directory, however, this will soon be automated with the use of bots.

## Checklist

### DIFF
--- a/mkdocs/gen_api_docs.py
+++ b/mkdocs/gen_api_docs.py
@@ -1,4 +1,4 @@
-#  Copyright © 2025 Bentley Systems, Incorporated
+#  Copyright © 2026 Bentley Systems, Incorporated
 #  Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
 #  You may obtain a copy of the License at


pyproject.toml
```diff
@@ -11,6 +11,7 @@ dependencies = [
     "evo-colormaps[aiohttp,notebooks]>=0.0.2",
     "evo-compute[aiohttp,notebooks]>=0.0.1rc2",
     "jupyter",
+    "mkdocstrings-python",
```
BenLewis-Seequent: Shouldn't this really be placed within the `dev` dependency group? Don't think users benefit much from having docs generator being installed.

GriffinBaxterSeequent: Fixed, thanks!

PR #209, opened 2026-02-17
Auth widget now only shows instances and workspaces.

davidknight-seequent:
## Description

The changes made in this PR are designed to simplify the Jupyter notebook widget that handles user auth.

## Changes
- The `organisation` menu has been renamed to `instance`.
- The `hub` menu is no longer shown. There is only ever 1 hub per instance so the hub selection is handled automatically under-the-hood.

NOTE: When the Evo discovery v3 endpoint is released we can revise/simplify the underlying logic of this widget.

## Checklist

### DIFF
--- a/packages/evo-sdk-common/pyproject.toml
+++ b/packages/evo-sdk-common/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "evo-sdk-common"
 description = "Python package that establishes a common framework for use by client libraries that interact with Seequent Evo APIs"
-version = "0.5.15"
+version = "0.5.16"
 requires-python = ">=3.10"
 license-files = ["LICENSE.md"]
 dynamic = ["readme"]


GriffinBaxterSeequent (approved):

AdamFreemanSeequent (approved):

packages/evo-sdk-common/src/evo/notebooks/widgets.py
```diff
@@ -264,7 +268,6 @@ def __init__(self, transport: ITransport, authorizer: IAuthorizer, discovery_url
             [
                 widgets.HBox([build_img_widget("EvoBadgeCharcoal_FV.png"), self._btn, self._loading_widget]),
                 widgets.HBox([self._org_selector]),
-                widgets.HBox([self._hub_selector]),
```
wordsworthc: It's probably worth removing the hub selector altogether, rather than just hiding it

davidknight-seequent: Good idea! I've removed it now if you're happy to take another look.

wordsworthc (approved):

PR #216, opened 2026-02-19
Typed objects block model

denis-simo:
This PR brings block models under the same typed object envolope as recently added objects.

Key parts of this pull requets:
* BLock models are now an OPTIONAL dependency for objects. **We want to have a single point of entry** and block models on evo are not just references from the perspective of the user - they are actual block models that the user expects to interact with. If the user does not install block models, they can interact with block model metadata handling the current dependency in a similar way to pandas and numpy. Evo-sdk user will get the whole package, sneaky developer who only cares about very thin set of changes - can get evo-objects with cut down dependencies.
* Block model dependencies were adjusted so that by default it can be used just to parse metadata. Interaction with data is now an optional dependency, this way for basic services that don't care about interacting with evo-data and that user evo-objects there is no extra dependency added.
* Block models support dataframe conversions, pretty printing via widgets, report creation, and automated mapping/creation from `object_from_reference/path`.
* I have shared typed definitions in Common, without going as far as adding shared functionality that relies on numpy to avoid new depenency on common.
* Block model in objects acts as a wrapper for a fully featured typed objects in evo-block models that can exist by itself.

* I have not added any converters to and from block model and 3D grid... but it is not that hard as interfaces are similar though not identical and because the objects are different enough - we do not want to have more shared interfaces at this stage other than common base blocks.

* OCTREE and FULLY SUBBLOCKED modes support was left out for now (via typed Objects) - PR is large already

This is a meaninful update, so I have bumped up the versions.  If we want to release this, I can write up the changelong now as well, though there is one more PR to come after this for Kriging compute.

## Checklist

### DIFF
--- a/packages/evo-objects/src/evo/objects/typed/block_model_ref.py
+++ b/packages/evo-objects/src/evo/objects/typed/block_model_ref.py
@@ -110,14 +110,9 @@ class BlockModel(BaseSpatialObject):
 
     block_model_version_uuid: Annotated[UUID | None, SchemaLocation("block_model_version_uuid")]
 
-    _geometry_raw: Annotated[BlockModelGeometry, SchemaLocation("geometry")]
+    geometry: Annotated[BlockModelGeometry, SchemaLocation("geometry")]
 
     _attributes_raw: Annotated[list[dict], SchemaLocation("attributes")] = []
-
-    @property
-    def geometry(self) -> BlockModelGeometry:
-        """The geometry definition of the block model."""
-        return self._geometry_raw
 
     @property
     def attributes(self) -> BlockModelAttributes:


RyanMillerSeequent (approved):
Have only looked at `packages/evo-objects` and everything LGTM there.

packages/evo-blockmodels/src/evo/blockmodels/typed/units.py
```diff
@@ -0,0 +1,175 @@
...
+
+    :param context: The context containing environment and connector.
+    :return: List of available units.
+
+    Example:
+        units = await get_available_units(manager)
+        for unit in units:
+            print(f"{unit.unit_id}: {unit.description} ({unit.symbol})")
+    """
+    from evo.blockmodels import BlockModelAPIClient
```
wordsworthc: Is there any reason this can't be a module-level import?

packages/evo-blockmodels/src/evo/blockmodels/__init__.py
```diff
@@ -10,7 +10,27 @@
...
+    BaseTypedBlockModel,
+    BoundingBox,
+    Point3,
+    RegularBlockModel,
+    RegularBlockModelData,
+    Size3d,
+    Size3i,
+    Units,
+    get_available_units,
+)
```
wordsworthc: Given that `pyarrow` and `pandas` are marked and optional, and that they are imported by `evo.blockmodels.typed`, I think it would be best not to export them from this namespace.

packages/evo-blockmodels/src/evo/blockmodels/typed/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+from .report import (
+    Aggregation,
+    MassUnits,
+    Report,
+    ReportCategorySpec,
+    ReportColumnSpec,
+    ReportResult,
+    ReportSpecificationData,
+)
+from .types import BoundingBox, Point3, Size3d, Size3i
```
wordsworthc: Even though these are now provided by `evo.common.typed`, I think it is still useful to export them from this namespace for convenience. Perhaps import them from `evo.common.typed` though

packages/evo-objects/src/evo/objects/typed/block_model_ref.py
```diff
@@ -0,0 +1,925 @@
...
+else:
+    _BLOCKMODELS_AVAILABLE = True
+
+
+def _require_blockmodels(operation: str = "This operation") -> None:
+    """Raise ImportError if evo-blockmodels is not installed."""
+    if not _BLOCKMODELS_AVAILABLE:
+        raise ImportError(
+            f"{operation} requires the 'evo-blockmodels' package. Install it with: pip install evo-objects[blockmodels]"
+        )
```
wordsworthc: I would guard the method definitions where `evo.blockmodels` is used, rather than raise an exception.

Do consider whether you expect any of the types in this module to be used _without_ `evo-blockmodels` installed.

packages/evo-sdk-common/src/evo/common/__init__.py
```diff
@@ -27,10 +27,15 @@
 )
 from .interfaces import IAuthorizer, ICache, IContext, IFeedback, ITransport
 from .service import BaseAPIClient
+from .typed import BoundingBox as BoundingBox
+from .typed import Point3 as Point3
+from .typed import Size3d as Size3d
+from .typed import Size3i as Size3i
```
wordsworthc: Why are these aliased?

```suggestion
from .typed import BoundingBox, Point3, Size3d, Size3i
```

packages/evo-blockmodels/src/evo/blockmodels/typed/base.py
```diff
@@ -0,0 +1,408 @@
...
+
+    async def list_reports(self, fb: IFeedback = NoFeedback) -> list[Report]:
+        """List all report specifications for this block model.
+
+        :param fb: Optional feedback interface for progress reporting.
+        :return: List of Report instances.
+        """
+        fb.progress(0.0, "Fetching reports...")
+
+        environment = self._client._environment
```
BenLewis-Seequent: Minor, you should be able to get the environment from the metadata:
```suggestion
        environment = self._metadata.environment
```

This avoids reaching for a private variable and access is more consistent with how you get the id of the block model.

packages/evo-objects/src/evo/objects/typed/block_model_ref.py
```diff
@@ -0,0 +1,902 @@
...
+        """Serialize this attribute as a target for compute tasks.
+
+        For existing attributes, returns an update operation referencing this attribute by name.
+
+        :return: A dictionary with operation type and reference.
+        """
+        return {
+            "operation": "update",
+            "reference": self.expression,
+        }
```
BenLewis-Seequent: Wonder if there should be a protocol or interface that define these methods, as `BlockModelPendingAttribute` define the same methods.

packages/evo-objects/src/evo/objects/typed/block_model_ref.py
```diff
@@ -0,0 +1,902 @@
...
+    :param attributes: List of attributes available on the block model.
+    :param coordinate_reference_system: Optional CRS for the block model.
+    """
+
+    block_model_uuid: UUID
+    block_model_version_uuid: UUID | None = None
+    geometry: BlockModelGeometry
+    attributes: list[BlockModelAttribute] = field(default_factory=list)
+
+    def compute_bounding_box(self) -> BoundingBox:
```
BenLewis-Seequent: Doesn't look like this handles rotations?

packages/evo-blockmodels/src/evo/blockmodels/__init__.py
```diff
@@ -14,3 +14,33 @@
...
+        Point3,
+        RegularBlockModel,
+        RegularBlockModelData,
+        Size3d,
+        Size3i,
+        Units,
+        get_available_units,
+    )
+except ImportError:
+    _UTILS_AVAILABLE = False
```
wordsworthc: `_UTILS_AVAILABLE` is unused. I would (again) advocate for not exporting members of `evo.blockmodels.typed` from the `evo.blockmodels` namespace, which would be more consistent with `evo.objects[.typed]`

packages/evo-objects/src/evo/objects/typed/block_model_ref.py
```diff
@@ -0,0 +1,902 @@
...
+        cell_data: "pd.DataFrame | None" = None
+        description: str | None = None
+        crs: str | None = None
+        size_unit_id: str | None = None
+        units: dict[str, str] = field(default_factory=dict)
+
+else:
+
+    @dataclass(frozen=True, kw_only=True)
+    class RegularBlockModelData:  # type: ignore[no-redef]
```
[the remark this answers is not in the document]
denis-simo: In theory you can have create a block model without cell data (and without Pandas). However I am not sure of a scenario where a user would actually install objects without pandas and block model withou pandas and just interact with the metadata..

I will remove the guard for similicity.  If a user eventuates that cares to have this unique trimmed down install and niche scenario - they can add the changes.

packages/evo-objects/src/evo/objects/typed/block_model_ref.py
```diff
@@ -0,0 +1,902 @@
...
+            >>> # After running kriging that adds attributes...
+            >>> block_model = await block_model.refresh()
+            >>> block_model.attributes  # Now shows the new attributes
+        """
+        # Refresh the typed block model delegate if it exists, so it's immediately up-to-date
+        if hasattr(self, "_typed_bm") and self._typed_bm is not None:
+            await self._typed_bm.refresh()
+        return await object_from_uuid(self._api_context, self.metadata.id)
+
+    if _BLOCKMODELS_AVAILABLE:
```
wordsworthc: Duplicated `if _BLOCK_MODELS_AVAILABLE:` block. Move the `refresh()` to before the first block

packages/evo-objects/src/evo/objects/typed/block_model_ref.py
```diff
@@ -0,0 +1,902 @@
...
+            "max_x": bbox.max_x,
+            "min_y": bbox.min_y,
+            "max_y": bbox.max_y,
+            "min_z": bbox.min_z,
+            "max_z": bbox.max_z,
+        }
+
+        return result
+
+    if _BLOCKMODELS_AVAILABLE:
```
wordsworthc: Another duplicated `if _BLOCK_MODELS_AVAILABLE:` block. Move the `_data_to_dict()` to before the first block

wordsworthc (approved):

PR #234, opened 2026-03-23
Cast large_string to string right before saving to Parquet.

davidknight-seequent:
## Description

Second attempt at getting rid of `large_string`.
- Reverted previous changes meant to fix this problem.
- Targeted functions that actually save dataframes to Parquet.

## Checklist

### DIFF
--- a/packages/evo-objects/src/evo/objects/utils/data.py
+++ b/packages/evo-objects/src/evo/objects/utils/data.py
@@ -203,7 +203,7 @@ class ObjectDataClient:
         columns_chunk_range = []
         for column in table.itercolumns():
             column = column.dictionary_encode()
-            if not (pa.types.is_string(column.type.value_type) or pa.types.is_large_string(column.type.value_type)):
+            if not pa.types.is_string(column.type.value_type):
                 raise TableFormatError("Category columns must be of type string")
             if not pa.types.is_int32(column.type.index_type):
                 # Currently, we only support int32 indices
@@ -231,8 +231,6 @@ class ObjectDataClient:
         )
 
         dictionary_values = all_chunks[0].dictionary
-        if pa.types.is_large_string(dictionary_values.type):
-            dictionary_values = pc.cast(dictionary_values, pa.string())
         lookup = pa.Table.from_arrays(
             [np.arange(len(dictionary_values), dtype=np.int32), dictionary_values], names=["key", "value"]
         )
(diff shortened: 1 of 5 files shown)


packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -300,7 +317,8 @@ def save_dataframe(
                 no table formats are specified, raised when the table does not match any known format.
             :raises StorageFileNotFoundError: If the destination does not exist or is not a directory.
             """
-            return self.save_table(pa.Table.from_pandas(dataframe), table_format=table_format)
+            table = _cast_large_strings(pa.Table.from_pandas(dataframe))
```
wordsworthc: I see your comment about `convert_dtypes`being rather slow, so I agree we shouldn't go that route in that case. Did you happen to try defining a `pa.Schema` to pass to `pa.Table.from_pandas()`? I suspect manipulating the schema before the dataframe is converted should generally be more performant.

If we add a schema handling hook in a similar to `_cast_large_strings` I think that would make it easier to support any other problematic and/or nice-to-have pydantic -> pyarrow conversions in the future.

The dataframe conversion might then look something like this:

```suggestion
            schema = _get_schema_from_dataframe(dataframe)
            table = pa.Table.from_pandas(dataframe, schema)
```

---
Sources:

* https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.from_pandas
* https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html#pyarrow.Schema.from_pandas

davidknight-seequent: Good call! Please check out the changes.

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -67,6 +67,29 @@ def _iter_refs(target: Any, _key: str | None = None) -> Iterator[str]:
...
+    :param dataframe: The pandas dataframe to get the schema from.
+    :return: A pyarrow schema with any configured type conversions applied.
+    """
+    schema = pa.Schema.from_pandas(dataframe)
+    fields = []
+    for field in schema:
+        if pa.types.is_large_string(field.type):
+            fields.append(field.with_type(pa.string()))
+        elif pa.types.is_dictionary(field.type) and pa.types.is_large_string(field.type.value_type):
+            fields.append(field.with_type(pa.dictionary(field.type.index_type, pa.string())))
```
wordsworthc: Where does `pa.dictionary`come into it? I didn't know we had any schemas that support it

rileyhowley-seequent (approved):

wordsworthc (approved):

PR #235, opened 2026-03-27
add support for simulation workflow task

geostats-BSY:
PR #235 is superseded by PR #267 and it will be closed.

Adds typed client support for the conditional simulation (turning bands) compute task, following the established kriging task pattern.

## Description

Add support for simulation workflow task, similar to the existing typed support for the kriging task. 
Additionally, included a formatter for pointset objects for a nice notebook view.

## Checklist

- [ x] I have read the contributing guide and the code of conduct

### DIFF
--- a/packages/evo-compute/docs/examples/simulation.ipynb
+++ b/packages/evo-compute/docs/examples/simulation.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "df76aa52",
+   "id": "0",
    "metadata": {},
    "source": [
     "# Conditional Simulation Compute Task\n",
@@ -16,7 +16,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "e5454a50",
+   "id": "1",
    "metadata": {},
    "source": [
     "## Authentication\n",
@@ -27,7 +27,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "101afbc7",
+   "id": "2",
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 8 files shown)


packages/evo-compute/docs/examples/simulation.ipynb
```diff
@@ -0,0 +1,692 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "variogram"
```
PaulCaygill-Seequent, line 116: Missing a pretty print comment or similar, looks unfinished in the full rendered layout IMO

Image

packages/evo-compute/src/evo/compute/tasks/simulation.py
```diff
@@ -0,0 +1,694 @@
...
+    """Mining cost per unit of waste material."""
+
+    mining_ore_cost: float
+    """Mining cost per unit of ore material."""
+
+    metal_recovery_fraction: float
+    """Fraction of metal expected to be recovered during processing."""
+
+
+class LossCalculation(BaseModel):
```
PaulCaygill-Seequent, line 200: Looks like there is now a module for this in `compute/tasks/geostatistics` (possibly due to when this PR was originally made), see https://github.com/SeequentEvo/evo-python-sdk/commit/8e5263629e1f63a58cb577bc5915631060b2ced2
Would that be fit for purpose here instead?

PaulCaygill-Seequent (approved):
Seems OK. 👍 
You may want to revisit the PR and factor in newer modules if they are relevant.

PR #236, opened 2026-03-27
Migrate widgets to anywidget

geostats-BSY:
**Summary**

Consolidates all notebook widgets into the `evo-widgets` package, adding modern anywidget-based interactive widgets (ServiceManagerWidget, FeedbackWidget) and enabling automatic HTML formatter registration on import.

**Changes**
New Features
* Interactive widgets — ServiceManagerWidget and FeedbackWidget now available from evo.widgets

Service manager widget:

image

Feedback Widget:

image

* Auto-registration — HTML formatters for typed objects register automatically when importing from evo.widgets in IPython/Jupyter (no %load_ext required)
Modern widget framework — Built on anywidget for improved cross-environment compatibility

**Summary**
Consolidates all notebook widgets into the evo-widgets package, adding modern anywidget-based interactive widgets (ServiceManagerWidget, FeedbackWidget) and enabling automatic HTML formatter registration on import.

**Changes**

* Interactive widgets — ServiceManagerWidget, FeedbackWidget available from evo.widgets
* Auto-registration — HTML formatters for typed objects register automatically when importing from evo.widgets in IPython/Jupyter (no %load_ext required)
* Modern widget framework — Built on anywidget for improved cross-environment compatibility

**Breaking Changes**
* `evo.notebooks` widgets are deprecated — imports still work but emit DeprecationWarning
* %load_ext evo.widgets is no longer required (but still supported)

**Migration**

Before: 
```
from evo.notebooks import ServiceManagerWidget, FeedbackWidget
%load_ext evo.widgets
```
After
```
from evo.widgets import ServiceManagerWidget, FeedbackWidget
# (auto-registers formatters)
```

## Checklist

- [ x] I have read the contributing guide and the code of conduct


packages/evo-compute/src/evo/compute/local-notebooks/kriging.ipynb
tim-hamblin-seequent, line 1: Did you mean to commit this file? It looks unrelated to the changes (unless I'm missing something here)

code-samples/geoscience-objects/simplified-object-interactions/simplified-object-interactions.ipynb
```diff
@@ -279,15 +269,14 @@
...
+   "id": "20",
    "metadata": {},
    "source": [
     "## Summary\n",
     "\n",
     "This notebook demonstrated the simplified workflow for working with Evo geoscience objects:\n",
     "\n",
     "1. **Authenticate** using `ServiceManagerWidget`\n",
-    "2. **Load the widgets extension** with `%load_ext evo.widgets`\n",
     "3. **Create objects** using typed data classes like `PointSetData`\n",
```
tim-hamblin-seequent: No more number 2?

packages/evo-objects/docs/examples/typed-objects-variograms.ipynb
```diff
@@ -605,21 +595,21 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "evo-sdk (3.10.18)",
```
tim-hamblin-seequent, line 598: What's the deal with this change? I didn't see this on any of the other notebook updates

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,206 @@
...
+    # string contains something that looks non-local, treat it as remote.
+    for value in candidate_hosts:
+        lower = value.lower()
+        if not any(local in lower for local in _LOCAL_HOSTNAMES):
+            return True
+
+    return False
+
+
+def _should_warn_insecure_notebook_usage() -> bool:
```
tim-hamblin-seequent: This function does nothing other than wrap another function... is it really needed?

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,206 @@
...
+        if token_str is None:
+            return None
+
+        try:
+            token_dict = json.loads(token_str)
+            return oauth.AccessToken.model_validate(token_dict)
+
+        except Exception:
+            # The token is invalid.
+            raise ValueError(f"Invalid token found in the environment file {_TOKEN_KEY}: {token_str!r}")
```
tim-hamblin-seequent: Is it safe to spit out the value from the .env file into an error message? I feel like it might not be best practice to potentially expose these kinds of sensitive values.
(I guess this is only for dev environments so maybe it doesn't matter?)

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,206 @@
...
+
+    _env: _OAuthEnv[_T_token]
+
+    async def reuse_token(self) -> bool:
+        """Attempt to reuse an existing token from the environment file.
+
+        :returns: True if a token was found and reused, False otherwise.
+        """
+        async with self._mutex:
+            if (token := self._env.get_token()) is None:
```
tim-hamblin-seequent: Did you mean to use `self._get_token` here?

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def set_token(self, token: _T_token | None) -> None:
+        if token is None:
+            new_value = None
+        else:
+            new_value = token.model_dump_json(by_alias=True, exclude_unset=True)
+        self.__dotenv.set(_TOKEN_KEY, new_value)
+
+
+class _NotebookAuthorizerMixin(_BaseAuthorizer[_T_token]):
```
tim-hamblin-seequent: What is the purpose of this being a mixin instead of just defining these methods directly on `AuthorizationCodeAuthorizer`?

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,206 @@
...
+                " notebooks in a private environment."
+            )
+        super().__init__(oauth_connector=oauth_connector, redirect_url=redirect_url, scopes=scopes)
+        self._env = _OAuthEnv(env)
+
+    @contextlib.asynccontextmanager
+    async def _unwrap_token(self) -> AsyncIterator[oauth.AccessToken]:
+        # Overrides the parent implementation so that we can automatically login at startup.
+        async with self._mutex:
+            if (token := self._env.get_token()) is None:
```
tim-hamblin-seequent, line 106: Again: did you mean to use `self._get_token`?

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def _get_token(self) -> _T_token | None:
+        return self._env.get_token()
+
+    def _update_token(self, new_token: _T_token) -> None:
+        super()._update_token(new_token)
+        self._env.set_token(new_token)
+
+
+class AuthorizationCodeAuthorizer(_NotebookAuthorizerMixin[oauth.AccessToken], oauth.AuthorizationCodeAuthorizer):
```
tim-hamblin-seequent: This name is a little confusing considering we already have an `AuthorizationCodeAuthorizer`

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,206 @@
...
+    async def _unwrap_token(self) -> AsyncIterator[oauth.AccessToken]:
+        # Overrides the parent implementation so that we can automatically login at startup.
+        async with self._mutex:
+            if (token := self._env.get_token()) is None:
+                token = await self._handle_login(timeout_seconds=60)
+                self._update_token(token)
+            yield token
+
+    async def refresh_token(self) -> bool:
+        succeeded = await oauth.AuthorizationCodeAuthorizer.refresh_token(self)
```
tim-hamblin-seequent: I'm guessing we can't call `super()` here because of the MRO, but if you didn't use a mixin you totally could 😉

packages/evo-widgets/src/evo/widgets/_interactive/env.py
```diff
@@ -0,0 +1,51 @@
...
+        self._file = cache.root / filename
+        self._file.touch(exist_ok=True)
+        self._cache = dotenv.dotenv_values(self._file, encoding="utf-8")
+
+    def get(self, key: str, default: str | None = None) -> str | None:
+        logger.debug(f"Getting environment variable: {key}")
+        return self._cache.get(key, default)
+
+    def set(self, key: str, value: str | None) -> None:
+        logger.debug(f"Updating environment variable: {key}={value!r}")
```
tim-hamblin-seequent: Again, I know this is only for local dev purposes, but logging a value you are setting in a .env feels sketchy to me.

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
```diff
@@ -0,0 +1,635 @@
...
+                # Check if selected is still valid
+                new_value = self.UNSELECTED[1]
+                for _, val in new_options:
+                    if val == selected:
+                        new_value = selected
+                        break
+
+            self.value = new_value
+            self._on_selected(new_value if new_value != self.UNSELECTED[1] else None)
+            self.disabled = len(new_options) <= 1
```
tim-hamblin-seequent: Will `new_options` ever be `<1`?

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
```diff
@@ -0,0 +1,635 @@
...
+    UNSELECTED = ("Select Workspace", _NULL_UUID)
+
+    def __init__(self, env: DotEnv, manager: ServiceManager, hub_selector: HubSelectorWidget) -> None:
+        self._manager = manager
+        super().__init__("Workspace", env)
+        hub_selector.observe(self._on_hub_selected, names="value")
+
+    async def refresh_workspaces(self) -> None:
+        self.loading = True
+        self.disabled = True
```
tim-hamblin-seequent: This will already be set by `_on_hub_selection`

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
```diff
@@ -0,0 +1,635 @@
...
+                await authorizer.login(timeout_seconds=timeout_seconds)
+
+    async def login(self, timeout_seconds: int = 180) -> ServiceManagerWidget:
+        """Authenticate the user and obtain an access token.
+
+        :param timeout_seconds: The maximum time (in seconds) to wait for the authorisation process to complete.
+        :returns: The current instance of the ServiceManagerWidget.
+        """
+        await self._service_manager._transport.open()
+        self.main_loading = True
```
tim-hamblin-seequent: You repeat this pattern of setting these flags, setting up a `try finally` and executing some code. It would be a good use case to use a `contextmanager` here. It would made `refresh_services` look a lot nicer instead of the double `try`

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
```diff
@@ -0,0 +1,635 @@
...
+        """Get an environment with the currently selected organization, hub, and workspace."""
+        return self._service_manager.get_environment()
+
+    def get_org_id(self) -> UUID:
+        """Gets the ID of the currently selected organization."""
+        return self._service_manager.get_org_id()
+
+    def get_cache(self) -> ICache:
+        """Gets the cache for this context."""
+        return self._cache
```
tim-hamblin-seequent, line 545: This is the same as the `cache` property

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
```diff
@@ -0,0 +1,635 @@
...
+def _generate_evo_url_from_object_reference(object_reference: str, view: str = "overview") -> str:
+    """Generate an Evo URL from a geoscience object reference URL."""
+    parsed = urlparse(object_reference)
+
+    hostname_parts = parsed.hostname.split(".") if parsed.hostname else []
+    if len(hostname_parts) < 1:
+        raise ValueError(f"Invalid object reference URL: cannot extract hub_id from hostname '{parsed.hostname}'")
+    hub_id = hostname_parts[0]
+
+    if "integration" in (parsed.hostname or ""):
```
tim-hamblin-seequent: What is this actually checking, I've never seen this style of if statement before. I don't think it's super clear what's actually being checked here.

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
```diff
@@ -0,0 +1,635 @@
...
+    if hasattr(value, "metadata") and hasattr(value.metadata, "url"):
+        return str(value.metadata.url)
+
+    if hasattr(value, "url"):
+        return str(value.url)
+
+    raise TypeError(f"Cannot serialize object reference of type {type(value)}")
+
+
+def _generate_evo_url_from_object_reference(object_reference: str, view: str = "overview") -> str:
```
tim-hamblin-seequent: It's not always easy to review this kind of logic. Setting up a few unit tests to show what it inputs and outputs would be great here.

packages/evo-widgets/src/evo/widgets/static/feedback.css
tim-hamblin-seequent, line 1: Would be good to have some constants setup for some of these hardcoded values. e.g. the colours.

packages/evo-widgets/src/evo/widgets/static/service_manager.js
```diff
@@ -0,0 +1,198 @@
...
...5QKfMTmCSzEqiY8IaKmZ/AZJiVQAOENzTA/ASaZVYCDRHe0BDzE2iKWQk0RnhDY8xPoHpmJdAo4Q2NMj+B6piVQOOENzTO/ASqYFYCEyC8YQLMT6BYZiUwIcIbJsT8BIphVgITJLxhgsxPYFRmJTBRwhsmyvwEwpmVwMQJb5g48xPIzqwEeEF4Ay+Yn0AWZiXAK8IbeMX8BAZjVgJcILyBC8xPYGdmJcBawhtYy/wEtmJWAlxJeANXMj+Ba5mVABsR3sBGzE/gArMSYCvCG9iK+Qm8YFYCbE14A1szP2HCzEqAnQlvYGfmJ0yIWQmwN+EN7M38hMaZlQCDEN7AIMxPaJBZCTAo4Q0MyvyEBpiVAFkIbyAL8xMqZVYCZCO8gWzMT6iIWQmQnfAGsjM/oWBmJUAY4Q2EMT+hMGYlQCjhDYQyP6EAZiXAKIQ3MArzE0ZgVgKMSngDozI/IYhZCTA64Q2MzvyEjMxKgGIIb6AY5icMyKwEKI7wBopjfsKezEqAIglvoEjmJ+zArAQomvAGimZ+wgbMSoAqCG+gCuYnrGFWAlRDeAPVMD/hHLMSoDrCG6iO+cmkmZUA1RLeQLXMTyblLP3bDrMSoFrCG6jabL64keJbgLdJcAPNEN5AE1KA300BboJSP8ENNEd4A82ZzRd9gN8X4FXqN9z3T5cnj6Z+IYD2CG+gWbP54nY6AX/Pt1y8/ikljwQ30DLhDTQvPQWlPwW/49suzpM0Jzma+oUA2ie8gclIzwG/lyLcjZjjOUuPg3zgOdzAlAhvYHLSjZirGcq7fgLCLNMNk4/cMAlMkfAGJm02XxykAL/tFDwbcxJg8jrhDfDSuVPwu27GHIQ5CcAbhDfAG9IWfBXhpijb8XQSgDWEN8AV0hTldvolwi/Xn24/TqfbT0v8gAAlEN4AGzp3Et4/nvB91617lm6WfOxmSYDrCW+AHaUX9NydWIQ73QbYkfAG2NO5GzNvNfx0lGV6Db/TbYAdCW+AgaWT8FvpV+278MN0s6RHAQLsSXgDZJR24bfO/ZoXfr37KclRmpM43QYYkPAGCJRmKX2AH6T/vFlAjB+n2D5ysg2Qj/AGGFmK8YP0axXmXYYX+fRPIfk8RXb/UpunbpAEiCO8AQqX5io3z33KW9d84j6uXwW1U2yAAnRd9/8BD1MC3v0+g/kAAAAASUVORK5CYII=";
```
tim-hamblin-seequent: I'm no js expert so I can't comment much on these files, but would adding these resources as actual files to the repo be a bit cleaner? Especially to reduce the duplication of the loading GIF?

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
```diff
@@ -0,0 +1,635 @@
...
+    def get_cache(self) -> ICache:
+        """Gets the cache for this context."""
+        return self._cache
+
+    def create_client(self, client_class: type[T_client], *args: Any, **kwargs: Any) -> T_client:
+        """Create a client for the currently selected workspace."""
+        return self._service_manager.create_client(client_class, *args, **kwargs)
+
+
+def _serialize_object_reference(value: Any) -> str:
```
tim-hamblin-seequent: This function relies super heavily on duck typing and has a super broad input type... Do we actually know what types we might be getting here? We could maybe use a `singledispatch` to handle this more cleanly?

packages/evo-widgets/src/evo/widgets/static/dropdown.css
```diff
@@ -0,0 +1,49 @@
+/* Dropdown Widget Styles
+ * Based on Seequent Web Design System
```
tim-hamblin-seequent, line 2: Can we see some screenshots on the PR of the new widgets?

tim-hamblin-seequent:
Looks like there are tests/linting that needs to be fixed up as well

packages/evo-compute/src/evo/compute/local-notebooks/kriging.ipynb
geostats-BSY: oops..

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
geostats-BSY: prob best not to.. thanks

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
geostats-BSY: def doesn't need to be, just mimicking the code in evo-sdk-common

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
geostats-BSY: noted!

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
geostats-BSY: hmm will give it a go

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,206 @@
...
+    # If the user has explicitly configured a public host, assume remote.
+    # These env vars are not standardised across all Jupyter frontends, but
+    # are used in several deployments; we only treat them as remote when the
+    # host clearly isn't a localhost alias.
+    candidate_hosts = []
+    for var in ("JUPYTERHUB_BIND_URL", "JUPYTER_SERVER_URL", "JUPYTER_BASE_URL"):
+        value = env.get(var)
+        if not value:
+            continue
+        candidate_hosts.append(value)
```
PaulCaygill-Seequent:
```suggestion
    candidate_hosts = [
        env_value
        for var in ("JUPYTERHUB_BIND_URL", "JUPYTER_SERVER_URL", "JUPYTER_BASE_URL")
        if (env_value:= env.get(var)) is not None
    ]
```

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,206 @@
...
+
+    # A very small and conservative hostname check: if the configured URL
+    # string contains something that looks non-local, treat it as remote.
+    for value in candidate_hosts:
+        lower = value.lower()
+        if not any(local in lower for local in _LOCAL_HOSTNAMES):
+            return True
+
+    return False
```
PaulCaygill-Seequent: I think the logic should probably be a bit more clear here. Our goal is to check that we have _don't_ have one of the default Jupyter server values set to a remote value, because that could cause issues that we should warn about. Additionally, we only have two values to check against, so we can simplify it a lot

```suggestion
    # check that all host variables use localhost in some way
    for host in candidate_hosts:
        if "localhost" in host or "127.0.0.1" in host:
            continue
        else: # host uses a remote URL
            return True
            
    return False
```

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,206 @@
...
+        """Attempt to reuse an existing token from the environment file.
+
+        :returns: True if a token was found and reused, False otherwise.
+        """
+        async with self._mutex:
+            if (token := self._env.get_token()) is None:
+                return False
+
+            if token.is_expired:
+                self._env.set_token(None)
```
PaulCaygill-Seequent: should this use `self._update_token` to None or does that not work because of the `super` call?

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
```diff
@@ -0,0 +1,635 @@
...
+            value=self.UNSELECTED[1],
+            disabled=True,
+        )
+        self.observe(self._on_value_change, names="value")
+
+    def _get_options(self) -> list[tuple[str, T]]:
+        raise NotImplementedError("Subclasses must implement this method.")
+
+    def _on_selected(self, value: T | None) -> None:
+        ...
```
PaulCaygill-Seequent: should this also raise NotImplemented?

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
```diff
@@ -0,0 +1,635 @@
...
+
+    pass
+
+
+class ServiceManagerWidget(anywidget.AnyWidget, IContext, metaclass=_ServiceManagerWidgetMeta):
+    """Main authentication and service discovery widget.
+
+    This is a modern anywidget-based implementation that provides authentication,
+    organization/hub/workspace selection, and API client creation.
+    """
```
PaulCaygill-Seequent: The metaclass section doesn't (to my mind) make sense. The widget does extend the anywidget, and it _implements all of the_ `IContext` interface. Why does it need the metaclass of IContexts metaclass? it isn't an interface anymore, is it? and if you wanted it to be, why not just explicitly use the Interface type?

If there's a reason, I'd like to see it added to the comment on the metaclass 🤞

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
PaulCaygill-Seequent: if we're extending the IContext interface we can probably keep these methods and drop the properties. Most code that uses IContext objects should be using these methods

PaulCaygill-Seequent (commented):
Just going to publish the few comments I have so far, as you seem to be actively working on this (and I've already had to delete one or two). I'd suggested a few tweaks to make code clearer - the PR is quite large though so I've only been able to cover a little bit with limited context, take it with a grain of salt

packages/evo-widgets/src/evo/widgets/static/service_manager.js
geostats-BSY: If I remember right anywidget doesn't permit serving the image directly. Could store the image then read and translate to base64 as needed to serve it client side... but not sure that is better?

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
geostats-BSY: yeah, i think problematic with the `super()`

packages/evo-widgets/src/evo/widgets/static/dropdown.css
geostats-BSY: yes, yes you can! but you may be dissapointed they look much like the old widgets. Will add to the PR after cleaning up re comments etc

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
geostats-BSY: much of this is duplicated anyway. adios confusing logic

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
geostats-BSY: moving this logic over to urls.py , will review more carefull there.

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
geostats-BSY: sure can add it.

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
geostats-BSY: was just navigating around this bit: `E   TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases`. 

Likely there is a better way

packages/evo-widgets/src/evo/widgets/_interactive/authorizer.py
```diff
@@ -0,0 +1,205 @@
...
+        :param oauth_connector: The OAuth connector to use for authentication.
+        :param redirect_url: The local URL to redirect the user back to after authorisation.
+        :param scopes: The OAuth scopes to request.
+        :param env: The environment to store the OAuth token in.
+        """
+        if _should_warn_insecure_notebook_usage():
+            warnings.warn(
+                "The evo.widgets.AuthorizationCodeAuthorizer is not secure, and should only ever be used in Jupyter"
+                " notebooks in a private environment."
+            )
```
denis-simo: I can see the workaround forthe prototype that I added to remind myself mostly.
If you would rather the warning does not appear, just cut is and extend Readme appropriately. Local environments are not really that much more secure either, all depends on the ports opened on the hsot.

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
```diff
@@ -0,0 +1,635 @@
...
+        super().__init__("Organisation", env)
+
+    def _get_options(self) -> list[tuple[str, UUID]]:
+        return [(org.display_name, org.id) for org in self._manager.list_organizations()]
+
+    def _on_selected(self, value: UUID | None) -> None:
+        self._manager.set_current_organization(value)
+
+
+class HubSelectorWidget(DropdownSelectorWidget[str]):
```
denis-simo: HUB is no longer shown. It disappeared on the main branch a few weeks before PDAC demo.

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
denis-simo: Can't have references integration or test in public code base.

Also see `evo.widgets.urls.py` , it contains the logic already.

packages/evo-widgets/src/evo/widgets/static/dropdown.css
denis-simo: Should be stored in assets along with other styles.css .

packages/evo-widgets/src/evo/widgets/static/service_manager.js
denis-simo: Load local image asset in the client. Hardcoding in the contents is not the way.

packages/evo-widgets/src/evo/widgets/__init__.py
```diff
@@ -15,18 +15,32 @@
 Evo SDK objects in Jupyter notebooks.
 
 Usage:
-    In a Jupyter notebook, load the extension to enable rich HTML rendering:
+    Simply import the package to enable rich HTML rendering:
 
-        %load_ext evo.widgets
```
denis-simo, line 20: The old command does not just load the widgets, it loads HTML formatting for objects. 
If we are replicating the equivalent setup, then we would want `from evo import widgets`  - so auto format if the user imports widgets rather than a sepcific service manager widget.

packages/evo-widgets/src/evo/widgets/__init__.py
```diff
@@ -15,18 +15,32 @@
...
+
+    To disable auto-registration, set EVO_WIDGETS_NO_AUTO_REGISTER=1 before importing.
 
 Manual API:
     from evo.widgets import get_viewer_url_for_objects
 
     # View multiple objects together
     url = get_viewer_url_for_objects(manager, [pointset, grid])
+
+Interactive Widgets (requires `anywidget` and `traitlets`):
```
denis-simo: just add a new requirement, I don't think we are going to have services trying to download only parts of the package. Widgets are intended to be one turn key that requires everything it needs to get the job done in a pretty way.

packages/evo-widgets/README.md
```diff
@@ -158,11 +205,14 @@ viewer_url = get_viewer_url_from_reference(ref_url)
 
 ## How It Works
 
-When you run `%load_ext evo.widgets`, the extension registers HTML formatters with IPython using `for_type_by_name`. This approach:
+When you import from `evo.widgets` in an IPython/Jupyter environment, HTML formatters are automatically registered using `for_type_by_name`. This approach:
 
 1. **Avoids hard dependencies** — The widgets package doesn't import model classes directly
 2. **Works with all typed objects** — Formatters are registered for the base class, so all subclasses are covered
 3. **Lazy loading** — Formatters only activate when the relevant types are actually used
+4. **Zero configuration** — No need to manually run `%load_ext`
```
denis-simo: Just mention what it does, rather than the changelog of what it no longer has to do.

denis-simo (changes requested):

packages/evo-widgets/src/evo/widgets/_interactive/widgets.py
geostats-BSY: am I right to think we no longer need this in the ServiceMangerWidget? just select hub based on Org?

PR #250, opened 2026-04-29
Add update block model

tim-hamblin-seequent:
## Description

Exposed the update block model method.

## Checklist


GriffinBaxterSeequent (approved):

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -992,6 +993,27 @@ async def get_deltas_for_block_model(
...
         )
 
+    async def update_block_model(
+        self,
+        bm_id: UUID,
+        update_block_model: UpdateBlockModel,
+    ) -> BlockModel:
+        """Update a block model's metadata.
+
+        Updates the block model name, description, and/or size unit ID for the given block model.
```
JamieZamodioSeequent: **suggestion (non-blocking):** `coordinate_reference_system` and `fill_subblocks` can also be set via this endpoint (see [API specs](https://developer.seequent.com/docs/api/blockmodel/operations#update-a-block-models-metadata), [the guide](https://developer.seequent.com/docs/guides/blockmodel/general-usage/block-model-metadata/update-block-model-metadata) is a bit behind unfortunately), can we update the docstring the reflect that?

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -992,6 +993,27 @@ async def get_deltas_for_block_model(
             delta_request_data,
         )
 
+    async def update_block_model(
```
JamieZamodioSeequent: **suggestion (non-blocking):** To avoid confusion with the other update block model endpoint (which updates columns), would it be better to rename this (and other similar code in this PR) to `update_block_model_metadata`?

JamieZamodioSeequent (commented):
Leaving a few minor questions first but otherwise looks good! Will try to have somebody else from BullHorn to have a look in case I missed something

packages/evo-blockmodels/tests/test_block_model_service_client.py
```diff
@@ -255,6 +266,48 @@ async def test_query_block_model_to_cache_no_cache(self) -> None:
...
+        with self.transport.set_http_response(
+            200, block_model_response.model_dump_json(), headers={"Content-Type": "application/json"}
+        ):
+            result = await self.bms_client.update_block_model(
+                bm_id=bm_uuid,
+                update_block_model=UpdateBlockModel(name="Updated Name", description="Updated description"),
+            )
+        self.assertEqual(result.name, "Updated Name")
+        self.assertEqual(result.description, "Updated description")
+        self.assertEqual(result.id, bm_uuid)
```
JamieZamodioSeequent: **suggestion (non-blocking):** Can we add size unit ID, CRS, and fill subblocks to the payload and asserts?

packages/evo-blockmodels/src/evo/blockmodels/client.py
adowling-seequent: For consistency with the existing naming scheme, i think this should be more than a 'suggestion', this should be a 'change-request'

adowling-seequent (changes requested):
@JamieZamodioSeequent has a couple of suggestions, however the update_block_model function name rename suggestion i think should be done, as a blocker.

adowling-seequent (approved):
LGTM

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -992,6 +993,28 @@ async def get_deltas_for_block_model(
...
 
+    async def update_block_model_metadata(
+        self,
+        bm_id: UUID,
+        update_block_model: UpdateBlockModel,
+    ) -> BlockModel:
+        """Update a block model's metadata.
+
+        Updates the block model name, description, coordinate reference system,
+        size unit ID, and/or fill subblocks setting for the given block model.
```
JamieZamodioSeequent: **nitpick:** when referring to sub-blocks in prose, it ought to have a hyphen:
```suggestion
        size unit ID, and/or fill sub-blocks setting for the given block model.
```

packages/evo-blockmodels/tests/test_block_model_service_client.py
```diff
@@ -255,6 +266,58 @@ async def test_query_block_model_to_cache_no_cache(self) -> None:
 
         self.transport.assert_no_requests()
 
+    async def test_update_block_model(self) -> None:
```
JamieZamodioSeequent: **nitpick:**

```suggestion
    async def test_update_block_model_metadata(self) -> None:
```

JamieZamodioSeequent (approved):
Reviewed with @rohancrookbain-seequent, LGTU (with minor comments)! Thanks @tim-hamblin-seequent :)

PR #267, opened 2026-04-30
Add conditioned simulator (consim) task SDK client

amirabbas-jalali-bentley:
## Description

Add a typed SDK client for the **conditioned simulator (consim)** geostatistics compute task (`evo.compute.tasks.geostatistics.conditioned_simulator`), kept in sync with the current `core-compute-tasks` task implementation.

### Changes
- Add `ConSimParameters` with full type coverage (neighborhood, variogram, distribution + tail extrapolation, block discretization, loss calculation, validation reporting, etc.).
- Add the generic attribute **filter** support that the task now exposes: `filter` (target grid) and `source_filter`, using the shared `Filter` model in `evo.compute.tasks.common` (`Filter` / `FilterCondition` / `AllOfFilter` / `AnyOfFilter`).
- Add `probability_above_cutoff` and `mean_above_cutoff` parameters, plus the corresponding `probability_above_cutoff_attributes` / `mean_above_cutoff_attributes` result attributes.
- Register the task module in `geostatistics/__init__.py`.
- Add unit tests for parameter construction, filter/cutoff serialization, and result handling.

### How to Test
1. `from evo.compute.tasks.geostatistics.conditioned_simulator import ConSimParameters`
2. `from evo.compute.tasks.common import Filter, FilterCondition`
3. Construct parameters (incl. `filter` / `source_filter` and `*_above_cutoff`) and verify JSON serialization matches the Task API schema.

## Checklist

- [x] Unit tests added/updated
- [x] All tests pass
- [x] Code has been linted
- [ ] No breaking changes

### DIFF
--- a/packages/evo-compute/src/evo/compute/tasks/common/filters.py
+++ b/packages/evo-compute/src/evo/compute/tasks/common/filters.py
@@ -1,0 +1,162 @@
+#  Copyright © 2026 Bentley Systems, Incorporated
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#      http://www.apache.org/licenses/LICENSE-2.0
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""Generic attribute filter expressions shared across geostatistics compute tasks.
+
+Several tasks (e.g. ``kriging`` and ``conditioned_simulator``) allow restricting a
+computation to a subset of locations by filtering an attribute on the source or
+target object.  A filter is a tree of expressions: a single
+:class:`FilterCondition`, or a composite :class:`AllOfFilter` (AND) / :class:`AnyOfFilter`
+(OR) combining nested expressions.
+
+Example:
+    >>> from evo.compute.tasks.common import Filter, FilterCondition, AllOfFilter
+    >>>
+    >>> # Single condition: include only locations where 'domain' is LMS1 or LMS2
+    >>> f = Filter(where=FilterCondition(attribute=grid.attributes["domain"], operator="in", values=["LMS1", "LMS2"]))
+    >>>
+    >>> # Compound condition: 'domain' in {1, 2} AND 'grade' >= 0.5
+    >>> f = Filter(
+    ...     where=AllOfFilter(
+    ...         filters=[
+    ...             FilterCondition(attribute=grid.attributes["domain"], operator="in", values=[1, 2]),
+    ...             FilterCondition(attribute=grid.attributes["grade"], operator="greater_than_or_equal_to", threshold=0.5),
+    ...         ]
+    ...     )
+    ... )
+"""
+
+from __future__ import annotations
+
+from typing import Annotated, Literal
+
+from pydantic import BaseModel, Field, model_validator
+
+from .source_target import AttributeExpression
+
+__all__ = [
+    "AllOfFilter",
+    "AnyOfFilter",
+    "Filter",
+    "FilterCondition",
+    "FilterExpression",
+    "FilterOperator",
+]
+
+FilterOperator = Literal[
+    "in",
+    "not_in",
+    "equal",
+    "not_equal",
+    "greater_than",
+    "greater_than_or_equal_to",
+    "less_than",
+    "less_than_or_equal_to",
+]
+"""Supported filter operators.
+
+Membership operators (``"in"``, ``"not_in"``) require ``values``; all other
+(numeric comparison) operators require ``threshold``.
+"""
+
+_MEMBERSHIP_OPERATORS: frozenset[str] = frozenset(("in", "not_in"))
+
+
+class FilterCondition(BaseModel):
+    """A leaf filter condition applied to a single attribute of the filtered object.
+
+    Provide exactly one operator and the matching payload:
+
+    - ``"in"`` / ``"not_in"``: membership tests — pair with ``values`` (a list of
+      string category names or integer category keys).
+    - ``"equal"``, ``"not_equal"``, ``"greater_than"``, ``"greater_than_or_equal_to"``,
+      ``"less_than"``, ``"less_than_or_equal_to"``: numeric comparisons — pair with
+      ``threshold`` (a single float).
+
+    Example:
+        >>> FilterCondition(attribute=grid.attributes["domain"], operator="in", values=["LMS1"])
+        >>> FilterCondition(attribute=grid.attributes["grade"], operator="greater_than", threshold=0.5)
+    """
+
+    type: Literal["condition"] = "condition"
+    """Discriminator identifying this as a leaf filter condition."""
+
+    attribute: AttributeExpression
+    """The attribute on the filtered object to evaluate the condition against."""
+
+    operator: FilterOperator
+    """The filter operator. Use ``in``/``not_in`` with ``values``; numeric operators with ``threshold``."""
+
+    values: list[str | int] | None = None
+    """Category values to match. Required for ``in``/``not_in``. Strings for category names, ints for category keys."""
+
+    threshold: float | None = None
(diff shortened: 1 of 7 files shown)


packages/evo-compute/src/evo/compute/tasks/common/filters.py
```diff
@@ -0,0 +1,156 @@
...
+
+class Filter(BaseModel):
+    """A filter expression defining which locations to include in a computation."""
+
+    where: FilterExpression
+    """The root filter expression. A condition, or an ``all_of`` / ``any_of`` composite."""
+
+
+AllOfFilter.model_rebuild()
+AnyOfFilter.model_rebuild()
```
PaulCaygill-Seequent, line 162: Hmmmm. No comment noting why, but I'll take a guess:
🔗 [from pydantic](https://pydantic.dev/docs/validation/latest/api/pydantic/base_model/#pydantic.BaseModel.model_rebuild)
> This may be necessary when one of the annotations is a ForwardRef which could not be resolved during the initial attempt to build the schema, and automatic rebuilding fails.

I'll post this as one review, but I can't say I'm a fan of this model structure.

packages/evo-compute/src/evo/compute/tasks/common/filters.py
```diff
@@ -0,0 +1,156 @@
...
+                    f"Operator '{self.operator}' requires 'values' (list of category names or integer keys)."
+                )
+            if self.threshold is not None:
+                raise ValueError(f"Operator '{self.operator}' does not accept 'threshold'.")
+        else:
+            if self.threshold is None:
+                raise ValueError(f"Operator '{self.operator}' requires 'threshold' (a numeric value).")
+            if self.values is not None:
+                raise ValueError(f"Operator '{self.operator}' does not accept 'values'.")
+        return self
```
PaulCaygill-Seequent, line 118: The way validation works here, feels like you actually want two different classes: a membership filter, and a threshold/numeric filter.

PaulCaygill-Seequent (commented):
I'd like you to take another look at the Filter section. The model rebuild calls at the end felt like an indicator that we should rework the module to use inheritance - perhaps incorporate the `abc` "Abstract Base Class" library in a way that lets us have useful types before we write the pydantic models.

A simple example I can think of, would be to three have abstract base classes that you group all of your implementations under:
- CollectedFilter
    - AllOfFilter
    - AnyOfFilter
- ThresholdFilter
    - GreaterThanFilter, LessThanFilter etc.
- MembershipFilter
    - InFilter
    - NotInFilter

This would then mean you could use the Base classes as your field typing (and make a Union for filter expression on the base classes as well) instead of having to rebuild the models.

Please consider the above options using inheritance here, and if you cannot/should not use it here for the models, let me know why

packages/evo-compute/tests/geostatistics/test_conditioned_simulator_tasks.py
```diff
@@ -0,0 +1,353 @@
...
+
+        mock_submit.assert_called_once()
+        _, kwargs = mock_submit.call_args
+        self.assertEqual(kwargs["topic"], "geostatistics")
+        self.assertEqual(kwargs["task"], "consim")
+        self.assertIsInstance(result, ConSimResult)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
PaulCaygill-Seequent: 👋

amirabbas-jalali-bentley:
> I'd like you to take another look at the Filter section. The model rebuild calls at the end felt like an indicator that we should rework the module to use inheritance - perhaps incorporate the `abc` "Abstract Base Class" library in a way that lets us have useful types before we write the pydantic models.
> 
> A simple example I can think of, would be to three have abstract base classes that you group all of your implementations under:
> 
> * CollectedFilter
>   
>   * AllOfFilter
>   * AnyOfFilter
> * ThresholdFilter
>   
>   * GreaterThanFilter, LessThanFilter etc.
> * MembershipFilter
>   
>   * InFilter
>   * NotInFilter
> 
> This would then mean you could use the Base classes as your field typing (and make a Union for filter expression on the base classes as well) instead of having to rebuild the models.
> 
> Please consider the above options using inheritance here, and if you cannot/should not use it here for the models, let me know why

Thanks for asking this great question, there were lots of discussion around the design of these filters and I agree, I should have captured a glimpse of them in the SDK.
With that said, I dug into this a bit more and I still think the `model_rebuild()` was the only way. `all_of/any_of` recursively contain filter expressions, so the field is inherently self-referential, and `model_rebuild()` is pydantic's documented pattern for recursive models.
Also pydantic does not polymorphically dispatch a base-class-typed field. If I type `filters: list[FilterBase]`, nested nodes validate as the base class and their payload is dropped (I verified this, nested in/greater_than nodes came back as empty base instances).
To parse nested nodes correctly you must type the field as a `Field(discriminator="type")` union, and a recursive discriminated union forward-references itself, which brings `model_rebuild()` straight back. Inheriting the leaf classes from `abc.ABC` doesn't change any of that, and since these are pure-data models with no methods, ABC mostly adds metaclass surface without buying the grouping benefit a plain base model wouldn't.

For splitting the leaf condition into per-operator classes it seems like a cleaner domain model and I'm open to it. But it's independent of `model_rebuild`, and it changes the discriminator/JSON-schema shape, which ripples into the discovery meta-schema, the deployed task's validation, and stored payloads. That's an upstream `core-compute-tasks` change with a migration, not something to fold into this SDK-mirror PR and it might not worth the cost ATM.

PaulCaygill-Seequent (approved):
@amirabbas-jalali-bentley Thanks for having another look for me.
I'd only looked at the method description directly and not at the broader docs ([here](https://pydantic.dev/docs/validation/2.3/usage/models/#rebuild-model-schema)) where pydantic specifically want you to use this tool. 
Your analysis makes a lot of sense too, I was focused on this code as "writing python to describe filters" and not what you seem to need from the use case, which is "unmarshalling recursive filter .json"

> pydantic does not polymorphically dispatch a base-class-typed field

There are possibly ways to do this using an Annotated field, kind of like how it's done in FastAPI endpoints, but as you point out later, possibly not worth it

> ABC mostly adds metaclass surface

yep, not worth using ABC for typing if we still have to lean on the recursive definition.

> For splitting the leaf condition [...] it changes the discriminator/JSON-schema shape, which ripples into the discovery meta-schema, the deployed task's validation

I guess that makes sense, and this is largely because the `type` field are based on an existing set of values, which we'd then need to change. hmmm, that's also unfortunate.

well given all of the above context I really like this PR! sorry for the noob review comment, it was quite a lot of required context that I was missing 😓  
The PR looks good overall, still really happy with the test coverage that you've been putting in.

amirabbas-jalali-bentley:
> @amirabbas-jalali-bentley Thanks for having another look for me. I'd only looked at the method description directly and not at the broader docs ([here](https://pydantic.dev/docs/validation/2.3/usage/models/#rebuild-model-schema)) where pydantic specifically want you to use this tool. Your analysis makes a lot of sense too, I was focused on this code as "writing python to describe filters" and not what you seem to need from the use case, which is "unmarshalling recursive filter .json"
> 
> > pydantic does not polymorphically dispatch a base-class-typed field
> 
> There are possibly ways to do this using an Annotated field, kind of like how it's done in FastAPI endpoints, but as you point out later, possibly not worth it
> 
> > ABC mostly adds metaclass surface
> 
> yep, not worth using ABC for typing if we still have to lean on the recursive definition.
> 
> > For splitting the leaf condition [...] it changes the discriminator/JSON-schema shape, which ripples into the discovery meta-schema, the deployed task's validation
> 
> I guess that makes sense, and this is largely because the `type` field are based on an existing set of values, which we'd then need to change. hmmm, that's also unfortunate.
> 
> well given all of the above context I really like this PR! sorry for the noob review comment, it was quite a lot of required context that I was missing 😓 The PR looks good overall, still really happy with the test coverage that you've been putting in.

@PaulCaygill-Seequent I really appreciate your eye for code and design quality, it gives me a fresh perspective to look at the design again and find aspects that can be improved and I enjoy explaining as it will help me fill the gaps in my knowledge as well.

Before I forget, I will also address the remaining concerns you brought up on this PR ASAP.

Thanks for putting the time and reviewing my PRs!

higordearaujo-bsy (approved):

GriffinBaxterSeequent (approved):

PR #270, opened 2026-05-03
Add PyPI compatibility test

GriffinBaxterSeequent:
## Description

This workflow simulates the experience of a developer installing our packages fresh from PyPI. Unlike the standard PR/push tests (which use the pinned uv.lock), a fresh install resolves the latest available versions of all dependencies, meaning any newly released package could introduce a breaking change that our pinned lockfile wouldn't catch.

## Checklist


AdamFreemanSeequent (approved):

.github/workflows/pypi-compatibility-test.yaml
```diff
@@ -0,0 +1,70 @@
...
+          - evo-colormaps
+          - evo-files
+          - evo-objects
+          - evo-compute
+          - evo-widgets
+        python-version:
+          - "3.10"
+          - "3.11"
+          - "3.12"
+          - "3.13"
```
davidknight-seequent, line 38: Can we add 3.14?

davidknight-seequent (changes requested):

.github/workflows/pypi-compatibility-test.yaml
GriffinBaxterSeequent: Sure, I added it to the other workflows too as it was missing from them.

GriffinBaxterSeequent: I take it back, there's a lot of issues with that, outside the scope of this PR

davidknight-seequent (approved):

PR #272, opened 2026-05-04
Tweaked a workflow.

davidknight-seequent:
## Description

Experimenting with adding Python 3.14 support to the compatility tests.

## Checklist


GriffinBaxterSeequent:
@davidknight-seequent this won't fail, as `pypi-compatibility-test.yaml‎` doesn't run on PRs, you may want to add 3.14 support to the other workflows

PR #274, opened 2026-05-05
Extending Python compatibility testing

davidknight-seequent:
### Changes

**CI Workflow (`pypi-compatibility-test.yaml`)**
- Reduced schedule from daily to weekly (Sundays)
- Removed the separate `resolve-latest-dependencies` job — each test job now deletes its own lockfile to resolve latest deps directly
- Added Python 3.14 to the test matrix
- Upgraded to `actions/checkout@v6`
- Added a new `run-notebook-tests` job that executes notebook tests with Evo credentials from secrets

**Notebook test framework (`code-samples/tests/`)**
- Cleared `AUTH_EXCLUDE_NOTEBOOKS` — all 6 previously excluded notebooks are now included in CI runs
- Added `sort_by_dependencies()` with a declarative `AUTH_NOTEBOOK_DEPENDENCIES` map so `download-pointset` runs after `publish-pointset`
- Extended `_PLATFORM_SPECIFIC_IMPORTS` to support Python version constraints, skipping `geosoft` on Python >= 3.14

**Notebook fixes**
- Updated `blockmodels/api-examples.ipynb`, `files/api-examples.ipynb`, and `files/sdk-examples.ipynb` to be CI-compatible

## Checklist


code-samples/tests/notebook_helpers.py
```diff
@@ -39,20 +39,16 @@
...
-    # Cascading failure: file_info set by an earlier interactive cell.
-    "code-samples/files/sdk-examples.ipynb",
-    # Intermittent 404: file upload may not propagate before the read-back poll.
-    "code-samples/files/api-examples.ipynb",
-    # geosoft is Windows-only; excluded on all platforms since CI matrix covers Windows separately.
-    "code-samples/geoscience-objects/publish-regular-2d-grid/publish-regular-2d-grid.ipynb",
-    # Requires a pre-existing pointset object ID to be set manually before running.
-    "code-samples/geoscience-objects/download-pointset/download-pointset.ipynb",
-]
+AUTH_EXCLUDE_NOTEBOOKS: list[str] = []
```
GriffinBaxterSeequent: I'd say get rid of this and it's usage to discourage this from ever needing to happen again 😁

.github/workflows/pypi-compatibility-test.yaml
```diff
@@ -53,18 +36,56 @@ jobs:
...
-          name: latest-deps-uv-lock
+      - name: Remove lockfile to resolve latest dependencies
+        shell: bash
+        run: rm -f uv.lock
 
       - name: Run tests against latest dependencies
         run: |
           uv run --package ${{ matrix.package }} pytest packages/${{ matrix.package }}/tests/
+
+  run-notebook-tests:
```
GriffinBaxterSeequent, line 55: I don't think we need `run-notebook-tests`. The workflow is simulating a fresh PyPI install, and code-samples isn't on PyPI. Users always get it from the repo with the pinned lockfile, so there's no "latest deps" scenario to validate. Notebook tests are already covered by the existing scheduled run-notebook-tests.yaml. Adding them here would just be extra CI minutes.

GriffinBaxterSeequent (changes requested):

.github/workflows/pypi-compatibility-test.yaml
davidknight-seequent: Many users can't or won't use `uv` which makes the lockfile redundant. Also, the notebooks represent more of a real world scenario because that is what most users are directly engaging with.

GriffinBaxterSeequent (approved):

AdamFreemanSeequent (approved):

PR #275, opened 2026-05-06
[Typed objects] Add support for SchemaLocation annotation on ClassVar variables

BenLewis-Seequent:
## Description

In preparation for supporting typed objects drillhole collections, this PR makes a couple of changes:
- Support for SchemaLocation annotation on ClassVar variables, which is used to represent constants that appear in the schemas. For example, the [type](https://github.com/SeequentEvo/evo-schemas/blob/8579a26e2df575241abe07cf86d92f1174ab2656/schema/objects/downhole-collection/1.3.1/downhole-collection.schema.json#L15-L19) property within the downhole collections schema.
- Fixed issue where properties and sub models on DataTableAndAttributes sub-classes where being ignored. For downhole collections, the distance tables include a [unit](https://github.com/SeequentEvo/evo-schemas/blob/8579a26e2df575241abe07cf86d92f1174ab2656/schema/components/distance-table/1.2.0/distance-table.schema.json#L31-L34) property which we want include the appropriate class, but this issue caused it to be difficult to expose.   

The proof of concept of the downhole collection object is here: https://github.com/SeequentEvo/evo-data-converters/pull/202

## Checklist

### DIFF
--- a/packages/evo-objects/src/evo/objects/typed/_data.py
+++ b/packages/evo-objects/src/evo/objects/typed/_data.py
@@ -143,4 +143,5 @@ class DataTableAndAttributes(SchemaModel):
         builder = SchemaBuilder(cls, context)
         await builder.set_sub_model_value("_table", table_df)
         await builder.set_sub_model_value("attributes", attr_df)
-        return builder.document
+
+        return await builder.build_from_data(data, skip_sub_models={"_table", "attributes"})
--- a/packages/evo-objects/src/evo/objects/typed/_data.py
+++ b/packages/evo-objects/src/evo/objects/typed/_data.py
@@ -144,4 +144,5 @@ class DataTableAndAttributes(SchemaModel):
         await builder.set_sub_model_value("_table", table_df)
         await builder.set_sub_model_value("attributes", attr_df)
 
+        # As we specially handle the _table and attributes sub-models above, we skip them in the default data processing of build_from_data.
         return await builder.build_from_data(data, skip_sub_models={"_table", "attributes"})
--- a/packages/evo-objects/src/evo/objects/typed/_model.py
+++ b/packages/evo-objects/src/evo/objects/typed/_model.py
@@ -275,10 +275,11 @@ class SchemaModel:
                 inner_args = get_args(annotation)
                 inner_type = inner_args[0]
                 base_type, schema_location, _ = _get_base_type(inner_type)
+                # Skip class variables without a SchemaLocation, e.g. data_columns: ClassVar[list[str]] = []
                 if schema_location is None:
                     continue
 
-                # Treat such ClassVar fields as constants, which also have a corresponding value in the schema
+                # Treat such remaining ClassVar fields as constants, which also have a corresponding value in the schema
                 type_adapter = TypeAdapter(base_type)
                 prop = SchemaConstantProperty(
                     jmespath_expr=schema_location.jmespath_expr,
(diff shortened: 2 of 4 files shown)


packages/evo-objects/src/evo/objects/typed/_model.py
```diff
@@ -234,6 +270,25 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
...
         for field_name, annotation in hints.items():
+            # Handle ClassVar fields
+            if get_origin(annotation) is ClassVar:
+                inner_args = get_args(annotation)
+                inner_type = inner_args[0]
+                base_type, schema_location, _ = _get_base_type(inner_type)
+                if schema_location is None:
+                    continue
+
+                # Treat such ClassVar fields as constants, which also have a corresponding value in the schema
```
daniel-kinney-seequent: How does it handle things like `data_columns: ClassVar[list[str]] = []`?

daniel-kinney-seequent (approved):

packages/evo-objects/src/evo/objects/typed/_model.py
BenLewis-Seequent: It is filtered out by the previous if condition. It only gets here if there is a `SchemaLocation` annotation like: `data_columns: ClassVar[Annotated[list[str], SchemaLocation("columns")]] = []`

julienpringotseequent (approved):

packages/evo-objects/src/evo/objects/typed/_data.py
```diff
@@ -143,4 +143,5 @@ async def _data_to_schema(cls, data: Any, context: IContext) -> Any:
         builder = SchemaBuilder(cls, context)
         await builder.set_sub_model_value("_table", table_df)
         await builder.set_sub_model_value("attributes", attr_df)
-        return builder.document
+
+        return await builder.build_from_data(data, skip_sub_models={"_table", "attributes"})
```
rileyhowley-seequent, line 148: I see we do `builder.set_sub_model_value` but on the `build_from_data` we skip the _table and attributes. Why is this?

packages/evo-objects/src/evo/objects/typed/_model.py
```diff
@@ -234,6 +270,26 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
 
         # Process the resolved annotations
         for field_name, annotation in hints.items():
+            # Handle ClassVar fields
+            if get_origin(annotation) is ClassVar:
+                inner_args = get_args(annotation)
+                inner_type = inner_args[0]
+                base_type, schema_location, _ = _get_base_type(inner_type)
+                # Skip class variables without a SchemaLocation, e.g. data_columns: ClassVar[list[str]] = []
```
rileyhowley-seequent: Curious as to why we are skipping class variables without a `SchemaLocation`

rileyhowley-seequent (approved):
Apologies for the delay. Just some questions around a couple of things that stood out to me most. Nothing to block

packages/evo-objects/src/evo/objects/typed/_model.py
BenLewis-Seequent: Because we are only looking for class variable/fields that will get reflected in the JSON objects. Other class variables continue to work as normal. The SchemaLocation annoation is used to indicate where in the object the constant value(in this case) should be produced, so without this it is assumed it shouldn't be reflected in the schema.

packages/evo-objects/src/evo/objects/typed/_data.py
BenLewis-Seequent: Because we have already populated the schema with the data above, which involves some special handling due to the data in the Geoscience object being split into several tables, but we want to accept a single table, with lots of columns. Thus the automatic building by build_from_data doesn't work.

rileyhowley-seequent (approved):

RyanMillerSeequent (approved):

PR #279, opened 2026-05-07
Update workspace sdk and notebooks

sachinseizer:
## Description

In this PR, i have added the following endpoints to the workspace SDK:

restore deleted workspace
get thumbnail
put thumbnail
delete thumbnail

In addition to these endpoints:
- updated example workspace-client jupyter notebook within evo-sdk-common package with all the new endpoints, as well as cleaned up the notebook to make it easier to work through
- added tests for the new endpoints, and added test coverage for a few existing endpoints that were missing it
- refactored test folder for workspaces to be easier to navigate
- updated evo.aio to support bytearray return type

## Checklist


Gibson-Seequent (approved):

packages/evo-sdk-common/src/evo/workspaces/client.py
```diff
@@ -499,3 +523,47 @@ async def update_instance_user_roles(self, user_id: UUID, roles: list[UUID]) ->
...
+            workspace_id=str(workspace_id),
+            additional_headers={"Accept": "image/jpeg, image/png"},
+        )
+
+    async def put_thumbnail(self, workspace_id: UUID, thumbnail: bytearray):
+        """
+        uploads the thumbnail image for a workspace.
+        :param workspace_id: The ID of the workspace to upload the thumbnail for.
+        :param thumbnail: The thumbnail image as a byte array.
+        :returns: The thumbnail image as bytes.
```
shivangj-10: docstring says "returns thumbnail as bytes" but the method returns `None`. Is the docstring correct?

packages/evo-sdk-common/src/evo/workspaces/client.py
```diff
@@ -87,6 +87,10 @@ async def get_service_health(self, check_type: HealthCheckType = HealthCheckType
         """
         return await get_service_health(self._connector, "workspace", check_type=check_type)
 
+    """
```
shivangj-10: Might be worth removing bare string literals in this file for consistency across the codebase.

packages/evo-sdk-common/tests/workspaces/test_workspace_client.py
```diff
@@ -1,621 +0,0 @@
...
-        with mock.patch("evo.workspaces.client.get_service_health") as mock_get_service_health:
-            await self.workspace_client.get_service_health()
-        mock_get_service_health.assert_called_once_with(self.connector, "workspace", check_type=HealthCheckType.FULL)
-
-    def _empty_content(self) -> str:
-        data = """{"results": [], "links": {"first": "http://firstlink", "last": "http://lastlink",
-                "next": null, "previous": null, "count": 0, "total": 0}}"""
-        return data
-
-    async def test_list_workspaces_default_args(self):
```
shivangj-10, line 178: I can see this test case with default arg is not included in `test_endpoints/test_workspaces.py` Is it expected?

packages/evo-sdk-common/pyproject.toml
shivangj-10: A version bump make sense?

shivangj-10 (changes requested):

packages/evo-sdk-common/tests/workspaces/test_workspace_client.py
sachinseizer: This test was duplicate coverage with test_list_workspaces, so I removed it.  same for test_list_workspaces_summary_default_args.

shivangj-10 (approved):
LGTM

PR #282, opened 2026-05-14
Update workspace SDK to use non-deprecated endpoints

keerynj04:
## Description

Regenerated the workspace SDK routes to update from deprecated user instance routes, mainly adding /users to the following routes:

- List instance users
- Add user to instance
- Remove instance user
- Update instance user roles

The associated tests were also updated to match the new routes.
## Checklist


packages/evo-sdk-common/src/evo/common/connector.py
```diff
@@ -201,6 +201,9 @@ async def call_api(
                 if key in specified_headers or isinstance(value, list):
                     raise RuntimeError(f"Multiple values not supported in header '{key}'")
                 else:
+                    # Remove duplicate headers
+                    if key.title() in headers:
+                        del headers[key]
```
sachinseizer: What was the problem that necessitated this change?

keerynj04: That was based off this [failing test](https://github.com/SeequentEvo/evo-python-sdk/actions/runs/25890418231/job/76428622334), but I think I have a better solution in mind so I'll look into that

wordsworthc: I don't think this is the best approach, and the failing tests are most likely a result of incorrect usage.

Please revert this change and see my comment re. authorization header in generated code.

packages/evo-sdk-common/src/evo/workspaces/client.py
```diff
@@ -87,6 +87,10 @@ async def get_service_health(self, check_type: HealthCheckType = HealthCheckType
         """
         return await get_service_health(self._connector, "workspace", check_type=check_type)
 
+    async def _get_authorization(self) -> str:
+        headers = await self._connector._authorizer.get_default_headers()
+        return headers.get("Authorization", "")
+
```
wordsworthc: Authorization headers are injected by `ApiConnector` - I think the best approach here is to fix the generated code to exclude the authorization parameter, eliminating the need for this.

Is there perhaps some schema defect that's causing the generated code to look this way? I don't recall it coming up before

wordsworthc (changes requested):

packages/evo-sdk-common/src/evo/workspaces/client.py
keerynj04: The schema's were changed to set the authorization to required, but this was intended for rule enforcement with other API endpoints using the schema. For the generated code, I think it would be best to create a filter that removes the added authorization section taken from the schema. This keeps the schema design correct, but does not generate the authorization section so it can be correctly injected by ApiConnector.

sachinseizer (approved):

wordsworthc (approved):
Thanks @keerynj04, this looks a lot better now

PR #263, opened 2026-04-30
Add continuous-distribution task SDK client

amirabbas-jalali-bentley:
## Description

Add typed SDK client for the **continuous-distribution** geostatistics compute task (`evo.compute.tasks.continuous_distribution`).

### Changes
- Add `ContinuousDistributionParameters` model with full type coverage
- Register task module in `tasks/__init__.py`
- Add unit tests for parameter construction and serialization

### How to Test
1. `from evo.compute.tasks.continuous_distribution import ContinuousDistributionParameters`
2. Construct parameters and verify JSON serialization matches the Task API schema

## Checklist

- [x] I have read the contributing guide and the code of conduct
- [x] Unit tests added/updated
- [x] All tests pass
- [x] Code has been linted
- [ ] No breaking changes

PR #289, opened 2026-05-21
Add declustering compute task SDK client

amirabbas-jalali-bentley:
## Summary

Adds an SDK client for the unified **declustering** compute task (`geostatistics/declustering`).

### Design

The task computes grid-based declustering weights by measuring each sample's influence on evaluation locations. Two estimation modes are exposed as module-level factory functions:

- **`idw(source, grid, target, neighborhood, *, power=2.0)`** — inverse-distance weighting; higher `power` gives nearer samples more weight
- **`knn(source, grid, target, neighborhood)`** — arithmetic-mean nearest-neighbour (equal weights); typically paired with `max_samples=1`

Both return a `DeclusteringParameters` instance, dispatched to `DeclusteringRunner` via the `TaskRegistry`.

A `@model_serializer` ensures `power=None` is serialised as `null` (not omitted) under `exclude_none=True`, so the server correctly interprets KNN mode.

### Changes

- **New**: `declustering.py` — `DeclusteringParameters`, `DeclusteringResult`, `DeclusteringRunner`, factory functions `idw()` and `knn()`
- **New**: `test_declustering_tasks.py` — comprehensive test suite (registration, serialization, result handling, runner execution, factory functions)
- **Updated**: `geostatistics/__init__.py` — import and re-export declustering module
- **Updated**: `tasks/__init__.py` — re-export `DeclusteringResult`, add to `__all__`

### Companion PR

Backend task implementation: https://github.com/seequent/core-compute-tasks/pull/737

## Checklist
- [x] Follows existing SDK patterns (Parameters, TaskRunner, Result wrapper)
- [x] Factory functions `idw()` / `knn()` for clear mode selection
- [x] `model_serializer` preserves `power: null` for KNN mode under `exclude_none=True`
- [x] Tests pass (`218 passed`)
- [x] Linting passes (ruff)

### DIFF
--- a/packages/evo-compute/src/evo/compute/tasks/geostatistics/__init__.py
+++ b/packages/evo-compute/src/evo/compute/tasks/geostatistics/__init__.py
@@ -15,7 +15,12 @@ This package groups all geostatistics-topic compute tasks:
 
 - **kriging** — Kriging interpolation
 - **break_ties** — Spatial tie-breaking
+- **declustering** — Grid-based declustering weights
 - **location_wise** — Per-location ensemble statistics
+- **loss_calculation** — Classification by expected economic loss
+- **normal_score** — Normal score transformation
+- **profit_calculation** — Classification by expected economic profit
+- **simulation_report** — Simulation result reporting
 
 Task modules are imported here to trigger runner registration with the
 central :class:`~evo.compute.tasks.common.runner.TaskRegistry`.
@@ -28,6 +33,7 @@ Example:
 """
 
 from . import break_ties as _break_ties_module  # noqa: F401
+from . import declustering as _declustering_module  # noqa: F401
 from . import kriging as _kriging_module  # noqa: F401
 from . import location_wise as _location_wise_module  # noqa: F401
 from . import loss_calculation as _loss_calculation_module  # noqa: F401
@@ -35,6 +41,7 @@ from . import normal_score as _normal_score_module  # noqa: F401
 from . import profit_calculation as _profit_calculation_module  # noqa: F401
 from . import simulation_report as _simulation_report_module  # noqa: F401
 from .break_ties import BreakTiesResult
+from .declustering import DeclusteringResult
 from .kriging import (
     BlockDiscretisation,
     KrigingResult,
@@ -45,6 +52,7 @@ from .location_wise import LocationWiseResult
 __all__ = [
     "BlockDiscretisation",
     "BreakTiesResult",
+    "DeclusteringResult",
     "KrigingResult",
     "LocationWiseResult",
     "RegionFilter",
--- a/packages/evo-compute/src/evo/compute/tasks/geostatistics/declustering.py
+++ b/packages/evo-compute/src/evo/compute/tasks/geostatistics/declustering.py
@@ -1,0 +1,365 @@
+#  Copyright © 2026 Bentley Systems, Incorporated
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#      http://www.apache.org/licenses/LICENSE-2.0
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""
+Declustering compute task client.
+
+This module provides typed models for running the declustering task
+(geostatistics/declustering).  The task computes grid-based declustering
+weights by measuring each sample's influence on evaluation locations.
+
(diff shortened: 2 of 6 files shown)


packages/evo-compute/src/evo/compute/tasks/geostatistics/__init__.py
```diff
@@ -15,7 +15,12 @@
 
 - **kriging** — Kriging interpolation
 - **break_ties** — Spatial tie-breaking
+- **declustering** — Grid-based declustering weights
 - **location_wise** — Per-location ensemble statistics
+- **loss_calculation** — Classification by expected economic loss
+- **normal_score** — Normal score transformation
+- **profit_calculation** — Classification by expected economic profit
+- **simulation_report** — Simulation result reporting
```
PaulCaygill-Seequent, line 23: @amirabbas-jalali-bentley You've got this change in both #289 and #263 which given they're the same shouldn't be too much of an issue... but still worth noting you may run into merge commit issues

packages/evo-compute/src/evo/compute/tasks/geostatistics/declustering.py
```diff
@@ -0,0 +1,365 @@
...
+    ...     target=Target.new_attribute(pointset, "declustering_weights"),
+    ...     neighborhood=SearchNeighborhood(
+    ...         ellipsoid=Ellipsoid(ranges=EllipsoidRanges(200, 150, 100)),
+    ...         max_samples=1,
+    ...     ),
+    ... )
+    >>> result = await run(context, params)
+"""
+
+from __future__ import annotations
```
PaulCaygill-Seequent, line 54: er, why is this imported?

packages/evo-compute/tests/geostatistics/test_declustering_tasks.py
```diff
@@ -0,0 +1,380 @@
...
+            "target": {
+                "reference": "ref",
+                "name": "target",
+                "schema_id": "s",
+                "attribute": {"reference": "ar", "name": "an"},
+            },
+        }
+        result = DeclusteringResultModel.model_validate(data)
+        self.assertIsInstance(result, DeclusteringResultModel)
+        self.assertEqual(result.message, "ok")
```
PaulCaygill-Seequent: Very nitpicky comment incoming

Seems a bit odd to test the `model_validate` method here, given that it's a pretty simple base model extension as far as I can tell (like the behavior should be guaranteed to do what we expect because it's the same as the library method?)
Is there any particular behavior that you're really looking for? something more specific with the unpacking perhaps?

PaulCaygill-Seequent (approved):
a couple of nitpick comments. Also re: my previous comment, I would be interested to hear if there is a specific tool you're using, as it would help me better understand if there are more situations we want to cater to with the public SDK. 

Otherwise, PR seems fine

packages/evo-compute/src/evo/compute/tasks/geostatistics/__init__.py
amirabbas-jalali-bentley: Thanks for the heads up! I will fix them as soon as I merge the other one

packages/evo-compute/src/evo/compute/tasks/geostatistics/declustering.py
amirabbas-jalali-bentley: Good question, actually I had the same question when I found it in the Kriging SDK (the one I used as template for all the other SDKs).
Based on my understanding, `from __future__ import annotations` postpones the evaluation of type hints, treating them as strings until they are needed at runtime (lazily you might say) and prevents Python from evaluating method annotations eagerly at class definition time.
For `@runtime_checkable` protocols (line 106) this matters because Python uses the annotations to perform structural `isinstance()` checks, so deferring evaluation avoids subtle issues if annotation types aren't fully resolved yet at module load and eliminates `NameError` exceptions.

higordearaujo-bsy (approved):

packages/evo-compute/src/evo/compute/tasks/geostatistics/declustering.py
PaulCaygill-Seequent: Huh, that's a pretty interesting reason!
I don't really know all that much about the `runtime_checkable` behavior so that's pretty neat. It took me a bit to wrap my head around why we would use the protocols themselves instead of using a subclass, but it does make sense when we're wanting to create modules that can be completely standalone.

> deferring evaluations avoids subtle issues if annotation types aren't fully resolved yet

...I need to wrap my head around this a bit better, but I think I understand what you're saying here. I feel like the previous solutions I've seen to annotations is to use string annotations (e.g. `def myfunction(arg:"MyType"):` where the type is usable/findable by the IDE) so the idea of using a different mechanism to defer it is kind of interesting. I guess the restriction of having to define a Protocol limits what you can use to solve the problems that are causing `NameErrors`!

Neat, thanks for the deep dive 🙇

GriffinBaxterSeequent (approved):

PR #299, opened 2026-06-08
Fix/page type bound and asyncio gather

sultan-alzoghaibi-seequent:
## Description

   Two fixes for the workspaces SDK:

   **1. Fix `Page` type bound**

   Removed `bound=ResourceMetadata` from the `_Metadata` TypeVar in `Page`. The bound was incorrectly
   restrictive,. This caused type errors when using `Page[Workspace]` since `Workspace` does not inherit
  from `ResourceMetadata`.

   **2. Refactor `list_all_workspaces` to use `asyncio.gather`**

   Replaced the sequential while loop with `asyncio.gather` to fetch all remaining pages concurrently
   after the first page. This reduces total fetch time when multiple pages exist.

   ## Checklist

### DIFF
--- a/packages/evo-sdk-common/src/evo/workspaces/client.py
+++ b/packages/evo-sdk-common/src/evo/workspaces/client.py
@@ -203,7 +203,7 @@ class WorkspaceAPIClient:
             deleted=deleted,
             filter_user_id=filter_user_id,
         )
-        page_read_coroutines = []
+        page_read_coroutines: list[Awaitable[Page[Workspace]]] = []
 
         for i in range(offset + limit, first_page.total, limit):
             page_read_coroutines.append(self.list_workspaces(
@@ -222,6 +222,8 @@ class WorkspaceAPIClient:
         workspaces = first_page.items() + [ws for page in remaining_pages for ws in page.items()]
 
         return sorted(workspaces, key=lambda x: x.display_name)
+
+
 
     async def list_workspaces_summary(
         self,
--- a/packages/evo-sdk-common/src/evo/workspaces/client.py
+++ b/packages/evo-sdk-common/src/evo/workspaces/client.py
@@ -12,7 +12,7 @@
 from __future__ import annotations
 
 import asyncio
-from typing import Literal, TypeAlias, Awaitable
+from typing import Awaitable, Literal, TypeAlias
 from uuid import UUID
 
 from pydantic import ValidationError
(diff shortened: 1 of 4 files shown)


packages/evo-sdk-common/src/evo/workspaces/client.py
```diff
@@ -186,31 +187,45 @@ async def list_all_workspaces(
...
-            if offset >= workspace_page.total:
-                break
+            ))
+
 
+        remaining_pages = await asyncio.gather(*page_read_coroutines)
+        workspaces = first_page.items() + [ws for page in remaining_pages for ws in page.items()]
+
+        # Note: this sort overrides any order_by passed to this method. This was here before
         return sorted(workspaces, key=lambda x: x.display_name)
```
sachinseizer: I get this was here before, but... this sorting override feels like it should not be here.  What's the point of being able to pass in all these sorting arguments just to not have the data returned sorted based on those parameters.  Kinda feels like a bug imo.

sachinseizer (commented):
One comment, otherwise looks good.

packages/evo-sdk-common/src/evo/workspaces/client.py
sultan-alzoghaibi-seequent: did this for backwards compatibility, 

```
        if order_by is None:
            return sorted(workspaces, key=lambda ws: ws.display_name)
        return workspaces
```


if someone is currently passing in a order_by that not display_name and expects a display_name ordering then Ig thats on them, bc this would break there expected return

packages/evo-sdk-common/src/evo/workspaces/client.py
```diff
@@ -186,30 +187,46 @@ async def list_all_workspaces(
...
-            offset += limit
-            if offset >= workspace_page.total:
-                break
+            ))
+
+
+        remaining_pages = await asyncio.gather(*page_read_coroutines)
+        workspaces = first_page.items() + [ws for page in remaining_pages for ws in page.items()]
+
+        if order_by is None:
```
sachinseizer: I think this is redundant?  Workspaces should have already ordered things correct according to the query params that were sent to it.

sachinseizer (approved):
Looks good, just need to fix CI/ID linting failures

packages/evo-sdk-common/src/evo/workspaces/client.py
wordsworthc: @sachinseizer thanks for calling this out - the client-side sort is a leftover defensive programming move from before the API supported sort. @sultan-alzoghaibi-seequent I think it should be safe to remove it entirely here.

sachinseizer (changes requested):
Need to change the order_by behaviour based on our convo, as well as fix the failing pipeline.

sachinseizer (approved):

wordsworthc (approved):

wordsworthc (approved):

PR #303, opened 2026-06-16
Typed downhole collection

daniel-kinney-seequent:
## Description

This PR introduces a typed "DownholeCollection" object. It is more complicated than the other examples, but it follows the same principals. The implementation and test coverage follows the example of `point_set`.

The most awkward point of design is how the column units are handled. The `DownholeCollectionData` dataclass deals mostly with pandas dataclasses, which don't particularly have a concept of column metadata. So, it's a challenge to figure out how the caller is supposed to attach metadata (such as "units"). The solution I went with is to expect the client to attach auxiliary metadata to pandas `DataFrame.attrs`, which would look like `{'attribute_description': {<column names>:  <AttributeDescription>}, ...}`.

## Checklist

### DIFF
--- a/packages/evo-objects/src/evo/objects/typed/_data.py
+++ b/packages/evo-objects/src/evo/objects/typed/_data.py
@@ -1,3 +1,14 @@
+#  Copyright © 2026 Bentley Systems, Incorporated
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#      http://www.apache.org/licenses/LICENSE-2.0
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
 from __future__ import annotations
 
 from typing import Annotated, Any, ClassVar
--- a/packages/evo-objects/src/evo/objects/typed/_data.py
+++ b/packages/evo-objects/src/evo/objects/typed/_data.py
@@ -111,7 +111,7 @@ class DataTableAndAttributes(SchemaModel):
         if attr_df is not None:
             await self.attributes.set_attributes(attr_df, fb=fb)
         else:
-            await self.attributes.clear()
+            self.attributes.clear()
 
     def validate(self) -> None:
         """Validate that all attributes have the correct length."""
--- a/packages/evo-objects/src/evo/objects/typed/attributes.py
+++ b/packages/evo-objects/src/evo/objects/typed/attributes.py
@@ -1,4 +1,4 @@
-#  Copyright © 2025 Bentley Systems, Incorporated
+#  Copyright © 2026 Bentley Systems, Incorporated
 #  Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
 #  You may obtain a copy of the License at
--- a/packages/evo-objects/src/evo/objects/typed/attributes.py
+++ b/packages/evo-objects/src/evo/objects/typed/attributes.py
@@ -11,7 +11,9 @@
 
 from __future__ import annotations
 
+import typing
 import uuid
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Annotated, Any
 from uuid import UUID
 
@@ -23,6 +25,7 @@ from evo.common.utils import NoFeedback, iter_with_fb
 from evo.objects import DownloadedObject
 from evo.objects.utils.table_formats import (
     BOOL_ARRAY_1,
+    DATE_TIME_ARRAY,
     FLOAT_ARRAY_1,
     INTEGER_ARRAY_1_INT32,
     INTEGER_ARRAY_1_INT64,
@@ -67,6 +70,8 @@ def _infer_attribute_type_from_series(series: pd.Series) -> str:
         return "category"
     elif pd.api.types.is_string_dtype(series):
         return "string"
+    elif pd.api.types.infer_dtype(series, skipna=True) in ["date", "datetime", "datetime64"]:
+        return "date_time"
     else:
         raise UnSupportedDataTypeError(f"Unsupported dtype for attribute: {series.dtype}")
 
@@ -76,7 +81,33 @@ _attribute_table_formats = {
     "integer": [INTEGER_ARRAY_1_INT32, INTEGER_ARRAY_1_INT64],
     "bool": [BOOL_ARRAY_1],
     "string": [STRING_ARRAY],
+    "date_time": [DATE_TIME_ARRAY],
 }
+
+
+@dataclass
+class AttributeDescription:
+    discipline: str = ""
+    type: str = ""
+    unit: str | None = None
+    scale: str | None = None
+    extensions: dict[str, typing.Any] | None = None
+    tags: dict[str, str] | None = None
+
+    def to_schema(self):
+        result = {
+            "discipline": self.discipline,
+            "type": self.type,
+        }
+        if self.unit:
+            result["unit"] = self.unit
+        if self.scale:
+            result["scale"] = self.scale
+        if self.extensions:
+            result["extensions"] = self.extensions
+        if self.tags:
+            result["tags"] = self.tags
+        return result
 
 
 class Attribute(SchemaModel):
@@ -158,7 +189,7 @@ class Attribute(SchemaModel):
             table_formats = _attribute_table_formats.get(attribute_type)
             attr_doc["values"] = await data_client.upload_dataframe(df, table_format=table_formats, fb=fb)
 
-        if attribute_type in ["scalar", "integer", "category"]:
+        if attribute_type in ["scalar", "integer", "category", "date_time"]:
             attr_doc["nan_description"] = {"values": []}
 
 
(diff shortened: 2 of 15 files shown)


packages/evo-objects/src/evo/objects/typed/_data.py
```diff
@@ -111,7 +111,7 @@ async def from_dataframe(self, df: pd.DataFrame, fb: IFeedback = NoFeedback) ->
         if attr_df is not None:
             await self.attributes.set_attributes(attr_df, fb=fb)
         else:
-            await self.attributes.clear()
+            self.attributes.clear()
```
daniel-kinney-seequent, line 125: I assume this is correct? I ran the typed unit tests and everything seemed happy.

packages/evo-objects/src/evo/objects/typed/attributes.py
```diff
@@ -76,9 +81,35 @@ def _infer_attribute_type_from_series(series: pd.Series) -> str:
...
+        }
+        if self.unit:
+            result["unit"] = self.unit
+        if self.scale:
+            result["scale"] = self.scale
+        if self.extensions:
+            result["extensions"] = self.extensions
+        if self.tags:
+            result["tags"] = self.tags
+        return result
```
daniel-kinney-seequent, line 110: This class is expected to be found in a dataframe's attrs. I'm not sure if it's the best approach for specifying unit, etc, but @BenLewis-Seequent and I discussed and thought it could be good enough.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -0,0 +1,321 @@
...
+
+    @classmethod
+    async def _data_to_schema(cls, data: pd.DataFrame, context: IContext) -> Any:
+        result = await super()._data_to_schema(data, context)
+        attr_desc = data.attrs.get("attribute_descriptions", {}).get("distance")
+        if attr_desc is not None:
+            result["unit"] = attr_desc.unit
+        else:
+            result["unit"] = None
+        return result
```
daniel-kinney-seequent, line 285: Getting the "unit" here is awkward. The "data" in this case is a pandas dataframe, and what I needed is a way to specify a getter function (grabbing data from it's `attrs` dictionary) instead of pointing at a path in the data object.

The awkwardness of this is perhaps an indication that maybe the data class DownholeCollectionData should wrap its tables with our own table wrappers to make everything more explicit. This relates to my comment in `attributes.py`.

packages/evo-objects/tests/typed/test_downhole_collection.py
```diff
@@ -0,0 +1,408 @@
...
+        properties=properties,
+        attributes=attributes,
+        collections=collections,
+        distance_unit="m",
+        description=description,
+        tags=tags,
+    )
+
+
+class TestDownholeCollection(TestWithConnector):
```
daniel-kinney-seequent, line 115: The tests basically mirror the point_set tests.

packages/evo-objects/src/evo/objects/typed/attributes.py
```diff
@@ -263,17 +302,27 @@ async def _upload_attributes_to_list(
 
         for col in df.columns:
             series = df[col]
-            attribute_type = _infer_attribute_type_from_series(series)
+            attribute_type = _infer_attribute_type_from_series(df[col])
```
BenLewis-Seequent, line 305: Is there a reason this changed? Though I don't think it matters.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -0,0 +1,304 @@
...
+            offset = self.holes.iat[i, 1]
+            count = self.holes.iat[i, 2]
+            collar = tuple(self.properties.loc[i, _COORDINATE_COLUMNS])
+            path_table = self.path[offset : offset + count]
+            bboxes.append(self._compute_hole_bounding_box(path_table, collar))
+
+        return BoundingBox.combine(bboxes)
+
+    @staticmethod
+    def _compute_bounding_box_np(
```
BenLewis-Seequent, line 121: Does this assume the desurvey algorithm is "trench"([the basic tangent algorithm](https://www.seequent.com/the-dark-art-of-drillhole-desurveying/))?

BenLewis-Seequent (approved):
Looks good to me.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -0,0 +1,304 @@
...
+
+HolePath: TypeAlias = pd.DataFrame
+HoleChunks: TypeAlias = pd.DataFrame
+HoleProperties: TypeAlias = pd.DataFrame
+HoleAttributes: TypeAlias = pd.DataFrame
+Depths: TypeAlias = pd.DataFrame
+
+
+@dataclass
+class DistanceCollection:
```
fzygou01, line 65: We need an `IntervalCollection `type alongside `DistanceCollection`. The schema supports `collection_type: "interval"` with a FLOAT_ARRAY_2 (from, to) table. Could you add `IntervalCollection(name, holes, interval_table) `with the same pattern?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -0,0 +1,304 @@
...
+    :param distance_unit: The distance unit for the `path` table and the `properties` x/y/z
+    .
+    :param coordinate_reference_system: Optional EPSG code or WKT string for the coordinate reference system.
+    :param description: Optional description of the object.
+    :param tags: Optional dictionary of tags for the object.
+    :param extensions: Optional dictionary of extensions for the object.
+    """
+
+    path: HolePath
+    holes: HoleChunks
```
fzygou01, line 95: Does the user need to provide pre-computed `holes`? If so, this could be error-prone. Maybe you could compute `holes `automatically from the contiguous `hole_id `groups in the path/collection DataFrames.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -0,0 +1,304 @@
...
+@dataclass
+class DistanceCollection:
+    name: str
+    holes: HoleChunks
+    distance_table: Depths
+    collection_type: str = "distance"
+
+
+@dataclass(kw_only=True, frozen=True)
+class DownholeCollectionData(BaseSpatialObjectData):
```
fzygou01, line 73: The DownholeCollection schema includes a `desurvey `field (e.g., "none", "tangential", "minimum_curvature"). Could this be added to both `DownholeCollectionData `and the typed object?

fzygou01 (changes requested):

packages/evo-objects/src/evo/objects/typed/attributes.py
daniel-kinney-seequent: No reason. It's leftover from an earlier attempt. It doesn't matter so I left it.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: Yes. That might not be an accurate assumption but the schema doesn't provide a way to specify.

edit - Actually, it looks like schema does cover that. I'm inclined to leave handling of the other desurvey methods for a future improvement.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: We do need this to include `IntervalCollection`, but the data source I'm working with (GEF) doesn't include interval data and so it would be premature on my end. There's a good change I will move on to AGS soon, which does have interval data, and that would be a better time to do that.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: Yes, I should add `desurvey`. I ignored it because it wasn't relevant for the GEF data source but I'll add it in.

daniel-kinney-seequent: Added.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: Yes, they do need to provide holes, because `hole_id` is not a column in the path and collection DataFrames.

packages/evo-objects/tests/typed/test_downhole_collection.py
```diff
@@ -0,0 +1,412 @@
+#  Copyright © 2025 Bentley Systems, Incorporated
```
GriffinBaxterSeequent:
```suggestion
#  Copyright © 2026 Bentley Systems, Incorporated
```

GriffinBaxterSeequent (approved):

l-macdonald (approved):

Bill-Johnson-Seequent (approved):
Approving on behalf of object-maintainers. Purely reviewed from the perspective of API relevant changes and found no issues. Can't and won't comment on other feature quality/correctness.

Please ensure all PR checks are passing before merging.

denis-simo (approved):
reviewed widgets

fzygou01 (approved):

sergiopeixoto-seequent (approved):

l-macdonald (approved):

BenLewis-Seequent (approved):

PR #300, opened 2026-06-11
Update kriging task SDK client to match current task implementation

amirabbas-jalali-bentley:
## Description

Update the typed SDK client for the **kriging** geostatistics compute task (`evo.compute.tasks.geostatistics.kriging`) to match the current `core-compute-tasks` task implementation.

The kriging task replaced its old single-attribute `region_filter` with a generic **attribute filter expression** applied to the source and/or target object. This PR brings the SDK client in line with that change.

### Changes
- Add a shared, generic filter model in `evo.compute.tasks.common` (`Filter`, `FilterCondition`, `AllOfFilter`, `AnyOfFilter`, `FilterExpression`, `FilterOperator`). A filter is a `Filter(where=...)` tree of conditions (membership `in`/`not_in` with `values`, or numeric comparisons with `threshold`) combined via `all_of` (AND) / `any_of` (OR).
- Replace `KrigingParameters.target_region_filter` (and the `RegionFilter` class) with `source_filter` and `target_filter`, serialised into `source.filter` / `target.filter` to match the task schema.
- Update exports (`evo.compute.tasks`, `evo.compute.tasks.common`, `evo.compute.tasks.geostatistics`) — `RegionFilter` is removed; `Filter` / `FilterCondition` are now exported.
- Update unit tests to cover the new filter model and parameter serialisation.
- Update the kriging sample notebook with a section documenting the optional `source_filter` / `target_filter` capability.

### How to Test
1. `from evo.compute.tasks import Filter, FilterCondition`
2. Construct `KrigingParameters(..., target_filter=Filter(where=FilterCondition(attribute=..., operator="in", values=[...])))`
3. Verify `params.model_dump()["target"]["filter"]["where"]` matches the Task API schema.

## Checklist

- [x] I have read the contributing guide and the code of conduct
- [x] Unit tests added/updated
- [x] All tests pass
- [x] Code has been linted
- [ ] No breaking changes (the deprecated `RegionFilter` is removed in favour of the generic `Filter`)

PR #307, opened 2026-06-19
Add KNN and IDW estimation task SDK clients

amirabbas-jalali-bentley:
## Description

Adds typed SDK clients for the two distance-based estimation compute tasks in the `geostatistics` topic.

### KNN (`packages/evo-compute/src/evo/compute/tasks/geostatistics/knn.py`)

- `KNNParameters` — source, target, neighbourhood, and optional `source_filter` / `target_filter` (`Filter` objects, serialised into the nested `source` / `target` dicts via a custom `model_serializer`).
- `KNNResultModel` / `KNNResult` — rich result object with `message`, `target_name`, `target_reference`, `attribute_name`, `schema`, `get_target_object()`, and `to_dataframe()`.
- `KNNRunner` — `TaskRunner` subclass registered as `geostatistics:knn`.

### IDW (`packages/evo-compute/src/evo/compute/tasks/geostatistics/idw.py`)

- `IDWParameters` — same shape as KNN plus a `power: float` field (distance-weighting exponent, must be `> 0`).
- `IDWResultModel` / `IDWResult` — same rich result API as KNN.
- `IDWRunner` — `TaskRunner` subclass registered as `geostatistics:idw`.

### Other changes

- `geostatistics/__init__.py` — imports both modules (to register runners) and re-exports `IDWResult` / `KNNResult`.
- `tests/geostatistics/test_knn_tasks.py` — 19 tests covering registration, parameter serialisation (including filter injection into source/target), result properties, and async runner.
- `tests/geostatistics/test_idw_tasks.py` — 24 tests covering the same areas plus `power` validation.

## Checklist

- [x] Code is linted (ruff check passes with no errors)
- [x] All tests pass (43/43)
- [x] No breaking changes introduced

### DIFF
--- a/packages/evo-compute/src/evo/compute/tasks/geostatistics/idw.py
+++ b/packages/evo-compute/src/evo/compute/tasks/geostatistics/idw.py
@@ -1,0 +1,236 @@
+#  Copyright © 2026 Bentley Systems, Incorporated
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""Inverse distance weighting (IDW) estimation compute task client.
+
+The IDW task estimates values at target locations using a weighted average of
+accepted neighbours, where each weight is inversely proportional to the distance
+raised to ``power``.  A higher power concentrates influence on the closest
+samples; ``power=2`` is the classic IDW setting.
+
+Example:
+    >>> from evo.compute.tasks import run, SearchNeighborhood, Ellipsoid, EllipsoidRanges
+    >>> from evo.compute.tasks.common import Source, Target
+    >>> from evo.compute.tasks.geostatistics.idw import IDWParameters
+    >>>
+    >>> params = IDWParameters(
+    ...     source=Source(object=pointset, attribute="locations.attributes[?name=='grade']"),
+    ...     target=Target.new_attribute(grid, "idw_grade"),
+    ...     neighborhood=SearchNeighborhood(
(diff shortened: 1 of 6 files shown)


packages/evo-compute/src/evo/compute/tasks/geostatistics/idw.py
```diff
@@ -0,0 +1,228 @@
...
+    """Optional filter to restrict estimation to a subset of the target object."""
+
+    @model_serializer(mode="wrap")
+    def _serialize(self, handler: SerializerFunctionWrapHandler) -> dict[str, Any]:
+        result = handler(self)
+        if self.source_filter is not None:
+            result["source"]["filter"] = self.source_filter.model_dump()
+        if self.target_filter is not None:
+            result["target"]["filter"] = self.target_filter.model_dump()
+        return result
```
PaulCaygill-Seequent, line 142: Hmmmm, I guess I don't quite understand why we need to remap this hear instead of at construction.
In previous PRs you've added logic that is there to handle incoming payloads, but this seems like it's there for outgoing payloads instead, which is fine, I think. That makes me think maybe you're using this to accept a payload or multiple payloads in multiple different forms. 
I guess my only thing with this, if the source/target already has a filter field, we're overriding them? which is not necessarily the behaviour I'd expect, but maybe you're fine with that?

I'll trust there's a good reason here, it would be nice to see this behaviour noted in the "target_filter" and "source_filter" doc string though 👍

PaulCaygill-Seequent (approved):
Nice! ✅ 

I had some comments about the filter logic, but looking at your tests, we _specifically look_ for the filters being injected on top of the original objects. I suspect this has something to do with the structure of the task being different from the structure of how we specify the data (neither being something within easy control of this SDK too) and we just need to inject it for the compute task and nothing else.
I also laughed a bit at seeing the places we need to use "almostEqual" because of floating point numbers, good catches!

Looks good to me! :shipit:

packages/evo-compute/src/evo/compute/tasks/geostatistics/idw.py
amirabbas-jalali-bentley: Good point, I've expanded the docstrings to explain the pattern.
The fields are marked `exclude=True` so they don't appear at the top level of the serialised payload, and the custom serialiser injects them into the nested `source["filter"]` / `target["filter"]` dicts. This is the same pattern already in `kriging.py` (merged in #300), but I should have made it explicit here too. Updated in the latest commit.

higordearaujo-bsy (approved):

PR #308, opened 2026-06-19
Add conditional turning-bands simulation task SDK client

amirabbas-jalali-bentley:
## Description

Adds a typed SDK client for the `conditional-turning-bands` compute task (geostatistics topic).

### New file: `packages/evo-compute/src/evo/compute/tasks/geostatistics/conditional_turning_bands.py`

`ConditionalTurningBandsParameters` covers all task parameters:
- `source` / `source_attribute` — conditioning point set and attribute JMESPath
- `target` — reference to the 3-D grid to simulate onto
- `distribution` — pre-computed non-parametric continuous distribution object (from the `continuous-distribution` task)
- `variogram_model` — variogram model for the spatial covariance step
- `neighborhood` (`SearchNeighborhood`) — search ellipsoid for both simulation and kriging conditioning
- `kriging_method` — `"simple"` (default) or `"ordinary"`
- `block_discretization` (`BlockDiscretization`, re-exported from `conditioned_simulator`) — sub-block support correction
- `number_of_lines` — 1–1000, default 500
- `realizations` — 1–100, default 1
- `random_seed` — for reproducibility
- `filter` / `source_filter` — optional `Filter` objects (require the `rfc-185-filters` preview flag)

`ConditionalTurningBandsResult` provides:
- `target_name`, `target_reference`, `schema` properties
- `simulations_attribute` — the ensemble attribute on the target grid (or `None`)
- `get_target_object()` / `to_dataframe()` async helpers

`ConditionalTurningBandsRunner` — `TaskRunner` subclass registered as `geostatistics:conditional-turning-bands`.

### Other changes

- `geostatistics/__init__.py` — imports the new module and re-exports `ConditionalTurningBandsResult`.
- `tests/geostatistics/test_conditional_turning_bands_tasks.py` — 29 tests covering runner registration, parameter serialisation (defaults, ranges, filter variants, composite filters), result properties, and async runner.

## Checklist

- [x] Code is linted (ruff check passes with no errors)
- [x] All tests pass (29/29)
- [x] No breaking changes introduced

### DIFF
--- a/packages/evo-compute/src/evo/compute/tasks/geostatistics/conditional_turning_bands.py
+++ b/packages/evo-compute/src/evo/compute/tasks/geostatistics/conditional_turning_bands.py
@@ -1,0 +1,271 @@
+#  Copyright © 2026 Bentley Systems, Incorporated
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""Conditional turning-band simulation (conditional-tbsim) compute task client.
+
+The conditional turning-bands task performs a conditional simulation of a
+numeric attribute onto a 3-D grid using a pre-computed continuous distribution.
+Unlike the full conditioned-simulator workflow, this task takes an existing
+distribution object produced by the ``continuous-distribution`` task and skips
+the internal normal-score transform; this makes it faster when the distribution
+is already available.
+
+Example:
+    >>> from evo.compute.tasks import run, SearchNeighborhood, Ellipsoid, EllipsoidRanges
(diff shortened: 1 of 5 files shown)


packages/evo-compute/src/evo/compute/tasks/geostatistics/conditional_turning_bands.py
```diff
@@ -0,0 +1,280 @@
...
+    filter: Filter | None = None
+    """Optional filter restricting simulation to a subset of target-grid locations.
+
+    Requires the ``rfc-185-filters`` API preview flag.
+    """
+
+    source_filter: Filter | None = None
+    """Optional filter restricting conditioning to a subset of the source data.
+
+    Requires the ``rfc-185-filters`` API preview flag.
```
PaulCaygill-Seequent: Should we maybe not reference internal RFC's in a public SDK?

PaulCaygill-Seequent (commented):
Seems fine, but we should probably not reference RFC's here, or am I wrong?

packages/evo-compute/src/evo/compute/tasks/geostatistics/conditional_turning_bands.py
amirabbas-jalali-bentley: Good catch, thanks! Done.

higordearaujo-bsy (approved):

GriffinBaxterSeequent (approved):

PR #309, opened 2026-06-19
Admin workspace info

shivangj-10:
## Description

- Fixed admin-workspace-info.ipynb notebook by removing Unicode char that is causing notebook test failure

## Checklist


davidknight-seequent (approved):

davidknight-seequent:
We decided to fix the unicode bug instead, so you can keep your emoji!

PR #316, opened 2026-07-22
add typed dh intervals

geostats-BSY:
## Description

A small addition to typed objects to support downhole intervals

## Checklist


packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,331 @@
+#  Copyright © 2025 Bentley Systems, Incorporated
```
amirabbas-jalali-bentley:
```suggestion
#  Copyright © 2026 Bentley Systems, Incorporated
```

packages/evo-objects/tests/typed/test_downhole_intervals.py
```diff
@@ -0,0 +1,330 @@
+#  Copyright © 2025 Bentley Systems, Incorporated
```
amirabbas-jalali-bentley:
```suggestion
#  Copyright © 2026 Bentley Systems, Incorporated
```

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,331 @@
...
+    def num_intervals(self) -> int:
+        """The number of intervals in this object."""
+        return self._from_to._intervals._table.length
+
+    # ------------------------------------------------------------------
+    # Schema creation
+    # ------------------------------------------------------------------
+
+    @classmethod
+    async def _data_to_schema(cls, data: DownholeIntervalsData, context: IContext) -> dict[str, Any]:
```
amirabbas-jalali-bentley: Would you explain why we need to re‑implements schema building here instead of using the declarative `DataLocation/sub‑model` pattern used by other objects?
I think the manual approach here duplicates base‑class logic and will silently drift if `BaseObject/BaseSpatialObject` changes.

packages/evo-objects/tests/typed/test_downhole_intervals.py
```diff
@@ -0,0 +1,330 @@
...
+        TestWithConnector.setUp(self)
+        self.environment = Environment(hub_url=BASE_URL, org_id=ORG.id, workspace_id=WORKSPACE_ID)
+        self.context = StaticContext.from_environment(
+            environment=self.environment,
+            connector=self.connector,
+        )
+
+    @contextlib.contextmanager
+    def _mock_geoscience_objects(self):
+        mock_client = MockClient(self.environment)
```
amirabbas-jalali-bentley, line 67: It seems like the schema shape is never truly validated.
The test `MockClient` just echoes back whatever `_data_to_schema` writes, so tests pass even if field placement doesn't match the real downhole-intervals schema.
I think, we should add a test_json-style assertion (there should be an example in `test_downhole_collection.py`) verifying the raw written document.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
geostats-BSY: thanks will adopt the sub-models

amirabbas-jalali-bentley (approved):

l-macdonald (approved):

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+
+    # --- schema properties ---
+    is_composited: Annotated[bool, SchemaLocation("is_composited")]
+
+    # --- sub-models ---
+    _start: Annotated[StartLocations, SchemaLocation("start"), DataLocation("intervals")]
+    _end: Annotated[EndLocations, SchemaLocation("end"), DataLocation("intervals")]
+    _mid_points: Annotated[MidLocations, SchemaLocation("mid_points"), DataLocation("intervals")]
+    _from_to: Annotated[FromToModel, SchemaLocation("from_to")]
+    _hole_id: Annotated[HoleIdCategory, SchemaLocation("hole_id"), DataLocation("intervals")]
```
daniel-kinney-seequent: The "sub-models" shouldn't have leading underscores, because they are pubic.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+    creation_schema_version = SchemaVersion(major=1, minor=3, patch=0)
+
+    # --- schema properties ---
+    is_composited: Annotated[bool, SchemaLocation("is_composited")]
+
+    # --- sub-models ---
+    _start: Annotated[StartLocations, SchemaLocation("start"), DataLocation("intervals")]
+    _end: Annotated[EndLocations, SchemaLocation("end"), DataLocation("intervals")]
+    _mid_points: Annotated[MidLocations, SchemaLocation("mid_points"), DataLocation("intervals")]
+    _from_to: Annotated[FromToModel, SchemaLocation("from_to")]
```
daniel-kinney-seequent: Opinion - "from_to" mirrors the schema, but to my eyes it's a pretty awkward name. Personally, I would be tempted to call this "intervals". It's part of the public interface so it's worth considering at least.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+    _table: Annotated[DepthIntervalsTable, SchemaLocation("start_and_end")]
+
+
+class HoleIdCategory(Category):
+    """Categorical hole identifier column."""
+
+    @classmethod
+    async def _data_to_schema(cls, data: pd.DataFrame, context: IContext) -> Any:
+        category_table = data[[_HOLE_ID_COL]].astype("category")
+        return await super()._data_to_schema(category_table, context=context)
```
daniel-kinney-seequent: Optional suggestion for improvement: The "hole_id" sub-model showed up exactly like this in `downhole_collection.py` as `DownholeCategory`. It would be a put that somewhere common and use it in both places.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+
+    @classmethod
+    async def _data_to_schema(cls, data: Any, context: IContext) -> list[dict[str, Any]]:
+        if data is not None:
+            attr_cols = [c for c in data.columns if c not in _ALL_REQUIRED_COLS]
+            data = data[attr_cols] if attr_cols else None
+        return await super()._data_to_schema(data, context)
+
+
+class FromToModel(SchemaModel):
```
daniel-kinney-seequent: `FromToModel` looks like it will be useful for `DownholeCollection` and also interim drilling for planned drillhole groups (which isn't a "typed object" yet but hopefully will be some day), so it would be great to put this in a common location.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+    """Schema model for the from_to component of a downhole intervals object."""
+
+    _intervals: Annotated[IntervalsWrapper, SchemaLocation("intervals"), DataLocation("intervals")]
```
daniel-kinney-seequent: `_intervals` should not have a leading underscore, because it's part of the public interface.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+    unit: Annotated[str | None, SchemaLocation("unit")]
+
+    @classmethod
+    async def _data_to_schema(cls, data: DownholeIntervalsData, context: IContext) -> Any:
+        result = await super()._data_to_schema(data, context)
+        if data.depth_unit is not None:
+            result["unit"] = data.depth_unit
```
daniel-kinney-seequent: It's weird to see "unit" needing special handling. There seems to be a quirk (bug?) where the "DataLocation" is being ignored for this. I was able to "fix" it by changing the "Data" class to have a "unit" field instead of a "unit_depth" field, but that's probably not what you want.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+
+    :param name: The name of the object.
+    :param intervals: DataFrame containing the interval data. Required columns:
+
+        * ``hole_id`` — hole identifier (string or Categorical)
+        * ``from`` — depth of the top of the interval
+        * ``to`` — depth of the base of the interval
+        * ``x_start``, ``y_start``, ``z_start`` — 3D start-point coordinates
+        * ``x_end``, ``y_end``, ``z_end`` — 3D end-point coordinates
+        * ``x_mid``, ``y_mid``, ``z_mid`` — 3D mid-point coordinates
```
daniel-kinney-seequent: I just wanted to call your attention to the fact that the "Data" class has 9 columns for all of start, mid, and end, which is different from the schema which has 3 tables of 3 columns each, one for start, mid, and end. I don't see a problem with it if you like it but I wasn't sure if that was your decision or Claude's.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+
+    @classmethod
+    async def _data_to_schema(cls, data: pd.DataFrame, context: IContext) -> Any:
+        return await super()._data_to_schema(data[_DEPTH_COLS], context)
+
+
+class IntervalsWrapper(SchemaModel):
+    """Wraps the 'intervals' sub-object inside 'from_to'."""
+
+    _table: Annotated[DepthIntervalsTable, SchemaLocation("start_and_end")]
```
daniel-kinney-seequent: These wrappers don't seem to be needed (this one, `StartLocations`, etc).

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+        if len(self.attributes) > 0:
+            attr_df = await self.attributes.to_dataframe(*keys, fb=fb)
+            parts.append(attr_df)
+
+        return pd.concat(parts, axis=1)
+
+    def validate(self) -> None:
+        """Validate the object, checking that all attribute lengths match num_intervals."""
+        super().validate()
+        self.attributes.validate_lengths(self.num_intervals)
```
daniel-kinney-seequent: It might be worth validating that all tables have the same lengths, because that's not enforced by the schema but it's the intended structure.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+    async def to_dataframe(self, *keys: str, fb: IFeedback = NoFeedback) -> pd.DataFrame:
+        """Get all interval data as a single DataFrame.
+
+        The returned DataFrame has the following columns, in order:
+
+        * ``hole_id`` — hole identifier
+        * ``from``, ``to`` — depth interval
+        * ``x_start``, ``y_start``, ``z_start`` — start-point coordinates
+        * ``x_end``, ``y_end``, ``z_end`` — end-point coordinates
+        * ``x_mid``, ``y_mid``, ``z_mid`` — mid-point coordinates
```
daniel-kinney-seequent: The double backtick "``" looks a bit unusual to me. Can you have a look at other docstrings and make sure the style is consistent?

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+    :param intervals: DataFrame containing the interval data. Required columns:
+
+        * ``hole_id`` — hole identifier (string or Categorical)
+        * ``from`` — depth of the top of the interval
+        * ``to`` — depth of the base of the interval
+        * ``x_start``, ``y_start``, ``z_start`` — 3D start-point coordinates
+        * ``x_end``, ``y_end``, ``z_end`` — 3D end-point coordinates
+        * ``x_mid``, ``y_mid``, ``z_mid`` — 3D mid-point coordinates
+
+        Any additional columns are uploaded as interval attributes.
```
daniel-kinney-seequent: The individual attributes can have their own units. See "AttributeDescription" in the comments in `downhole_collection.py`. I think your code does handle it implicitly because it uses `Attributes` for a base class. It's worth adding a test for that and explaining that somewhere in the docstring. It looks like I forgot to add a test for that in `downhole_collection.py`🙈. If copilot can't make sense of it I can suggest a test.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+from .types import BoundingBox
+
+__all__ = [
+    "DownholeIntervals",
+    "DownholeIntervalsData",
+]
+
+# ---------------------------------------------------------------------------
+# Column name constants
+# ---------------------------------------------------------------------------
```
daniel-kinney-seequent: I'll just note that these comments, and also the `# --- ` ones, are generally not helpful and they make it obvious that AI code generation was used. I'll leave it to you whether you care about that.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
```diff
@@ -0,0 +1,303 @@
...
+        * ``x_end``, ``y_end``, ``z_end`` — end-point coordinates
+        * ``x_mid``, ``y_mid``, ``z_mid`` — mid-point coordinates
+        * Any attribute columns
+
+        :param keys: Optional attribute keys/names to include.  If omitted, all
+            attributes are included.
+        :param fb: Optional feedback object to report download progress.
+        :return: A combined DataFrame of all interval data and attributes.
+        """
+        hole_id_df = await self._obj.download_category_dataframe("hole_id", column_names=[_HOLE_ID_COL])
```
daniel-kinney-seequent: This is unusual. I think it should be calling `self.hole_id.to_dataframe()`. I'll suggest a change for this.

packages/evo-objects/tests/typed/helpers.py
```diff
@@ -67,6 +67,27 @@ async def download_dataframe(self, data: dict, fb=None, **kwargs) -> pd.DataFram
     async def download_attribute_dataframe(self, data: dict, fb) -> pd.DataFrame:
         return self.mock_client.get_dataframe(data["values"])
 
+    async def download_category_dataframe(
+        self, category_info, fb=None, *, nan_values=None, column_names=None
+    ) -> pd.DataFrame:
```
daniel-kinney-seequent: This helper is unusually elaborate. I'm suggesting a change which makes it just like the one above.

packages/evo-widgets/src/evo/widgets/formatters.py
```diff
@@ -165,14 +165,27 @@ def _format_attributes_spec(attributes, rows) -> str:
...
 
-def _format_all_attribute_specs(obj, rows) -> None:
-    # Build datasets section - add as rows to the main table
-    sub_models = getattr(obj, "_sub_models", [])
-    for dataset_name in sub_models:
-        dataset = getattr(obj, dataset_name, None)
-        if dataset and hasattr(dataset, "attributes") and len(dataset.attributes) > 0:
-            # Build attribute rows
-            rows.append((f"{dataset_name}", _format_attributes_spec(dataset.attributes, rows)))
+def _append_attribute_rows(rows: list[tuple[str, str]], *objects: Any) -> None:
```
daniel-kinney-seequent: This change doesn't make sense to me. It looks like you needed to include attributes if the attributes are on the top level object. That bit makes sense. But I didn't understand some of the other stuff. The code for making sure that attributes don't show twice in the formatted text looks to be related to the change on line 211 in `format_downhole_collection`: `_append_attribute_rows(rows, obj, obj.location)`. But I don't understand why that change would have been made. I'm hoping this is a code generation issue, but LMK if there's something I didn't understand about you needed.

Assuming I got things right... I'm suggesting a change to simplify this. And also, I commented out a couple tests that didn't make sense anymore. (If you accept my changes you can clean those up.)

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
daniel-kinney-seequent: There's no need for action on this one - you can resolve before merging.

code-samples/geoscience-objects/publish-downhole-intervals/publish-downhole-intervals.ipynb
daniel-kinney-seequent, line 1: I think a better place for this notebook would be `code-samples/geoscience-objects/simplified-object-interactions`.

daniel-kinney-seequent (approved):

code-samples/geoscience-objects/publish-downhole-intervals/publish-downhole-intervals.ipynb
geostats-BSY: moved!

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
geostats-BSY: happy to remove.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
geostats-BSY: yeah, this is intentional - from a notebook user perspective imo it is nice to have all the data in a signle table.

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
geostats-BSY: ah, didn't know that - agree, that would be worthwhile. done!

packages/evo-objects/src/evo/objects/typed/downhole_intervals.py
geostats-BSY: not sure exactly what you had in mind but I've added a test [here](https://github.com/geostats-BSY/evo-python-sdk/blob/d21999c6fb7325d1c2d86e5abe0f02e5a084d25c/packages/evo-objects/tests/typed/test_downhole_intervals.py#L314), does this suffice?

daniel-kinney-seequent (approved):

PR #323, opened 2026-07-30
Add evo.common.crs with EpsgCode and parse_crs

sergiopeixoto-seequent:
## Description

Moves EpsgCode from evo.objects.typed.types into a new evo.common.crs module in evo-sdk-common, and adds parse_crs() to normalise CRS references (int, 'EPSG:1234', 'urn:ogc:def:crs:EPSG::1234', bare numeric strings, and OGC WKT). The 404000 / 'unspecified' / '' / None sentinels all map to None, so callers no longer have to special-case them. EpsgCode is re-exported from evo.objects.typed for backwards compatibility. The Pydantic _load_crs / _dump_crs serialisers and the CoordinateReferenceSystem alias stay in evo-objects, as they are schema-shape concerns. Bumps evo-sdk-common to 0.5.26 and raises the evo-objects floor accordingly.

## Checklist

### DIFF
--- a/packages/evo-objects/pyproject.toml
+++ b/packages/evo-objects/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "evo-objects"
 description = "Python SDK for using the Seequent Evo Geoscience Object API"
-version = "0.5.0"
+version = "0.5.1"
 requires-python = ">=3.10"
 license-files = ["LICENSE.md"]
 dynamic = ["readme"]
--- a/packages/evo-objects/pyproject.toml
+++ b/packages/evo-objects/pyproject.toml
@@ -10,7 +10,7 @@ authors = [
 ]
 
 dependencies = [
-    "evo-sdk-common[jmespath]>=0.5.8",
+    "evo-sdk-common[jmespath]>=0.5.26",
     "pydantic>=2,<3",
 ]
 
--- a/packages/evo-sdk-common/src/evo/common/crs.py
+++ b/packages/evo-sdk-common/src/evo/common/crs.py
@@ -1,0 +1,113 @@
+#  Copyright © 2026 Bentley Systems, Incorporated
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#      http://www.apache.org/licenses/LICENSE-2.0
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""Coordinate reference system (CRS) types and parsing helpers."""
+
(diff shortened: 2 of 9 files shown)


packages/evo-objects/pyproject.toml
GriffinBaxterSeequent, line 1: You'll want a new version of evo-objects too. Also note we'll release evo-sdk-common first, then the new evo-objects, so it'll contain the latest evo-sdk-common as a dependency.

sergiopeixoto-seequent: Addressed with https://github.com/SeequentEvo/evo-python-sdk/pull/323/changes/1f7f2a7cd5a0b313a141e730fe636b73b5ec80af

GriffinBaxterSeequent (approved):

packages/evo-sdk-common/src/evo/common/crs.py
```diff
@@ -0,0 +1,113 @@
...
+
+    :raises TypeError: If the value is not an ``int``, ``str`` or ``None``.
+    :raises ValueError: If an EPSG-shaped value is outside the valid range
+        [``MIN_EPSG_CODE``, ``MAX_EPSG_CODE``].
+    """
+    if value is None:
+        return None
+
+    if isinstance(value, bool):
+        raise TypeError(f"CRS value must be an int, str or None, got {type(value).__name__}")
```
peter-kipping-seequent, line 96: Why do we do this check?
Surely the check on line 101 below will catch this?
Is there a reason we need an early exit before the is int check below?

packages/evo-sdk-common/src/evo/common/crs.py
```diff
@@ -0,0 +1,113 @@
...
+
+    if isinstance(value, int):
+        return None if value == UNSPECIFIED_EPSG_CODE else EpsgCode(value)
```
peter-kipping-seequent, line 99: Just a random suggestion but should the __new__ in the epsg class also raise some sort of 'UNKNOWN_EPSG' code if you pass in the magic number and we try and catch that here instead of an explicit check?
Might make the actually constructor slightly more flexible. But honestly happy with this way, just figured id ask.

peter-kipping-seequent (approved):
Two minor suggestion and cant speak too much to the actual functionality but happy with it from an overview POV

packages/evo-sdk-common/src/evo/common/crs.py
sergiopeixoto-seequent: `bool` needs to be handled before `int` because Python defines `bool` as an `int` subclass. Without this guard it enters the integer branch, that we want to avoid.

packages/evo-sdk-common/src/evo/common/crs.py
sergiopeixoto-seequent: `EpsgCode` is an existing strict value type and `404000` is not a valid EPSG value; it is an input sentinel meaning “unspecified.” I prefer to keep that interpretation in `parse_crs`, where it is converted directly to `None`.

GriffinBaxterSeequent (approved):

peter-kipping-seequent (approved):

PR #325, opened 2026-08-05
Downhole collection intervals

sergiopeixoto-seequent:
## Description
Implemented typed interval and indexed DownholeCollection support in evo-objects, including polymorphic collection handling, hole-index validation, DataFrame read APIs, prefetching, attribute-description round-tripping, and static-context creation. The branch also bumps evo-objects to 0.6.2

## Checklist


l-macdonald (approved):

packages/evo-objects/src/evo/objects/typed/_model.py
```diff
@@ -450,15 +450,30 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
...
                 args = get_args(base)
                 if args:
-                    cls._item_type = args[0]
+                    item_type = args[0]
+                    if isinstance(item_type, type):
+                        cls._item_type = item_type
+                    elif "_resolve_item_type" not in cls.__dict__ or "_data_to_schema" not in cls.__dict__:
+                        raise TypeError(
+                            "SchemaList with a non-class item type must override _resolve_item_type() and _data_to_schema()"
+                        )
```
daniel-kinney-seequent, line 459: I'm a bit concerned about `cls._item_type` being unset in this branch. It means the base class has to be careful of how and whether to use `cls._item_type`. I wonder if there's a cleaner way to handle it.

Just a thought - what do you think about a new `SchemaUnion` class which handles all the details of polymorphism? I imagine the usage would look something like:

```
class DownholeCollectionTable(SchemaUnion[DownholeDistanceTable, DownholeIntervalTable]):
    @classmethod
    def _type_from_obj(obj_dict):
        if ...:
             return DownholeDistanceTable
        else if ...:
             return DownholeIntervalTable

class DownholeCollectionTables(SchemaList[DownholeCollectionTable]):
    pass
```
IMO that looks clean and easy to get right. I think it would tidy up the `DownholeCollectionTables` implementation. @BenLewis-Seequent might have an opinion. It's not essential for this PR as far as I'm concerned.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -301,3 +472,49 @@ class DownholeCollection(BaseSpatialObject):
     desurvey: Annotated[str | None, SchemaLocation("desurvey")]
 
     type: ClassVar[Annotated[str, SchemaLocation("type")]] = "downhole"
+
+    async def prefetch_collections(self, *names: str, include_location: bool = True, **kwargs: Any) -> None:
```
daniel-kinney-seequent: Is it possible to use explicit args instead of kwargs?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -301,3 +472,49 @@ class DownholeCollection(BaseSpatialObject):
...
+        """Prefetch data referenced by named collections and optionally location data."""
+        from evo.objects.typed._prefetch import collect_data_ids
```
daniel-kinney-seequent: Is the local import necessary? I didn't expect an import cycle with `_prefetch`.

packages/evo-objects/src/evo/objects/typed/_prefetch.py
```diff
@@ -0,0 +1,72 @@
...
+
+    visit(documents)
+    return ids
+
+
+async def prefetch_object_data(
+    obj: DownloadedObject,
+    *,
+    data_ids: Sequence[str] | None = None,
+    max_concurrent: int = 100,
```
daniel-kinney-seequent: I'm curious if there is any logic behind allowing 100 simultaneous downloads?

packages/evo-objects/src/evo/objects/typed/attributes.py
```diff
@@ -87,18 +88,23 @@ def _infer_attribute_type_from_series(series: pd.Series) -> str:
 
 @dataclass
 class AttributeDescription:
-    discipline: str = ""
-    type: str = ""
-    unit: str | None = None
+    discipline: str | None = None
+    type: str | None = None
+    unit: Any | None = None
```
daniel-kinney-seequent: Looking at the schema, it looks like `discipline` is not optional and `type` is also not optional. And I'm not sure `unit` should be relaxed without a good reason. Is it possible to keep the types as they were? Unless there's a reason that I didn't see.

That probably has consequences for the other changes in this class.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -75,13 +94,13 @@ class DownholeCollectionData(BaseSpatialObjectData):
 
     :param name: The name of the object.
     :param holes: A DataFrame describing which parts of `path` belong to which holes.
-            Columns: hole_id, offset, count
-    :param properties: DataFrame for the properties of the holes. The ith row corresponds to the ith element of `holes`.
+            Columns: hole_index, offset, count. ``hole_index`` is a code in ``properties.hole_id``.
+    :param properties: DataFrame for the properties of the holes, joined to ``holes`` by the categorical hole-id code.
```
daniel-kinney-seequent: `joined to ``holes`` by the categorical hole-id code.`

I'm concerned that the way `hole_id` and `hole_index` are associated might not be valid, because I'm not sure that `hole_index` is necessarily 0-indexed or contiguous. I made a few comments here or there about that same thing.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -95,23 +114,66 @@ class DownholeCollectionData(BaseSpatialObjectData):
     holes: HoleChunks
     properties: HoleProperties
     attributes: HoleAttributes | None
-    collections: list[DistanceCollection]
-    distance_unit: str | None
-    desurvey: str | None
+    collections: list[DownholeCollectionEntry]
+    distance_unit: str | None = None
+    desurvey: str | None = None
```
daniel-kinney-seequent: Do you have a reason for preferring these to have default arguments? When I thought about this originally, I left them as required arguments to force the caller to make a domain decision about them.

The same question applies to `DistanceCollection.unit` and `IntervalCollection.unit`, which IMO would be better as a required argument.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -95,23 +114,66 @@ class DownholeCollectionData(BaseSpatialObjectData):
...
+
+    @property
+    def hole_id_dtype(self) -> pd.CategoricalDtype:
+        """The categorical dtype whose codes are used by every ``hole_index``."""
+        hole_ids = self.properties["hole_id"]
+        if isinstance(hole_ids.dtype, pd.CategoricalDtype):
+            return hole_ids.dtype
+        return pd.CategoricalDtype(categories=sorted(hole_ids.dropna().unique()))
```
daniel-kinney-seequent: I have a few comments/questions about `hole_id_dtype`.

It seems like it's being used as a mapping to bridge `properties.hole_id` and `holes.hole_index`. The way it's being used looks pretty indirect, by eventually calling `collars["hole_id"].astype(self.hole_id_dtype).cat.codes`. Is it possible to use a more direct technique?

Based on how it's being used, I have a couple concerns. The concerns apply to usage in a few places, but I'll just leave one comment here:
1. I'm not sure it's a valid assumption that `hole_index` is contiguous and 0-indexed. I just reviewed the schema and it doesn't look like the schema makes that demand. Maybe it ought to though. Based on my understanding, they could be any integers, as long as they correspond to the integers that show up in the collection tables.
2. A pandas categorical dtype exposes dense, zero-based positional codes, whereas `category-data` is a lookup table without that stipulation. At a glance it looks like this category dtype is meant to match the schema's lookup table, and it could be confusing during downloading an already published object if there is a mismatch.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -67,6 +73,19 @@ class DistanceCollection:
...
     distance_table: Depths
     collection_type: str = "distance"
+    unit: str | None = None
+
+
+@dataclass
+class IntervalCollection:
+    name: str
+    holes: HoleChunks
+    interval_table: Intervals
```
daniel-kinney-seequent: If you want we could rename `interval_table` and `distance_table` to `table`, just to make the generic code a bit nicer. That's breaking the public API but I would be surprised if anyone is using this.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -95,23 +114,66 @@ class DownholeCollectionData(BaseSpatialObjectData):
...
+            )
+            self._validate_hole_chunks(collection.holes, len(table), require_coverage=False)
+
+    def _validate_hole_chunks(self, holes: HoleChunks, table_length: int, *, require_coverage: bool) -> None:
+        required = {"hole_index", "offset", "count"}
+        if missing := required - set(holes.columns):
+            raise ObjectValidationError(f"Hole chunks are missing columns: {sorted(missing)}")
+        indices = holes["hole_index"].astype(int)
+        valid = set(range(len(self.hole_id_dtype.categories)))
+        if not set(indices).issubset(valid):
```
daniel-kinney-seequent: Re-stating a question from before - I don't think `indices` has to be 0-indexed or contiguous, so I'm not sure this is correct.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -95,23 +114,66 @@ class DownholeCollectionData(BaseSpatialObjectData):
...
+            raise ObjectValidationError("hole_index must be a code in properties['hole_id'] categorical dtype")
+        if indices.duplicated().any():
+            raise ObjectValidationError("Each hole_index may occur only once in a holes table")
+        if require_coverage and set(indices) != valid:
```
daniel-kinney-seequent: Because of my above comment I'm not sure if this is valid either. Would it work to just compare the lengths at this point? Because you already showed it's a subset.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -95,23 +114,66 @@ class DownholeCollectionData(BaseSpatialObjectData):
...
+            raise ObjectValidationError("Location holes must cover every hole_id categorical code exactly once")
+        offsets = holes["offset"].astype(int)
+        counts = holes["count"].astype(int)
+        if (offsets < 0).any() or (counts < 0).any() or ((offsets + counts) > table_length).any():
+            raise ObjectValidationError("Hole chunk offsets and counts must be within the associated table")
+        non_empty = sorted(zip(offsets[counts > 0], counts[counts > 0], strict=True))
+        expected_offset = 0
+        for offset, count in non_empty:
+            if offset != expected_offset:
+                raise ObjectValidationError("Hole chunk ranges must cover the associated table exactly once")
```
daniel-kinney-seequent, line 189: I'm not 100% sure about either part of the assertion, at least not based on a literal interpretation of the schema.
- I'm not sure the hole chunk ranges have to cover the table. There could be reasons that data is convenient to push this way.
- I'm not sure it's a 1-to-1 relationship. For example, I can imagine pushing a CPT instrument through the same hole twice. I don't know if that' something that actually happens.

The comment applies to `_table_by_hole()` also.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -254,6 +316,21 @@ class DownholeLocation(SchemaModel):
...
+    async def to_dataframe(self, fb: IFeedback = NoFeedback) -> pd.DataFrame:
+        """Return collars with a categorical ``hole_id`` aligned to hole-index codes."""
+        parts = [
+            await self.hole_id.to_dataframe(fb=fb),
+            await self.coordinates.to_dataframe(fb=fb),
+            await self.distances.to_dataframe(fb=fb),
+        ]
+        if len(self.attributes):
+            parts.append(await self.attributes.to_dataframe(fb=fb))
+        return pd.concat(parts, axis=1)
```
daniel-kinney-seequent, line 383: Just an observation - do you think it's important for the `to_dataframe()` method to be symmetrical with the Data class? I don't  mind but it did stand out to me.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -67,6 +73,19 @@ class DistanceCollection:
     holes: HoleChunks
     distance_table: Depths
     collection_type: str = "distance"
+    unit: str | None = None
```
daniel-kinney-seequent: For consistency with elsewhere in this file, I would name this "distance_unit". Or better yet "depth_unit", which was used for https://github.com/SeequentEvo/evo-python-sdk/pull/316.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -267,10 +344,13 @@ class DistanceTableDistances(DataTableAndAttributes):
     @classmethod
     async def _data_to_schema(cls, data: pd.DataFrame, context: IContext) -> Any:
         result = await super()._data_to_schema(data, context)
+        unit = data.attrs.get("unit")
```
daniel-kinney-seequent: `DistanceCollection` and `IntervalCollection` now have `unit`, so there shouldn't be a need to embed the distance unit in the pandas dictionary anymore. Same comment for `IntervalTableFromTo`.

It was originally me who decided to put the attribute descriptions into the dataclass's `attrs`. I don't think it's all that great to start with so I'm keen to stop doing that where possible.

If we're concerned about backwards compatibility we have to keep doing it the old way, but I think we can away with dropping it.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -283,9 +363,100 @@ class DistanceTable(SchemaModel):
 class DownholeDistanceTable(DistanceTable):
     holes: Annotated[HoleChunksTable, SchemaLocation("holes"), DataLocation("holes")]
 
+    async def to_dataframe(self, fb: IFeedback = NoFeedback) -> pd.DataFrame:
+        return await self.distance.to_dataframe(fb=fb)
```
daniel-kinney-seequent: Not important - `def to_dataframe()` could be pushed into `DistanceTable`.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -301,3 +472,49 @@ class DownholeCollection(BaseSpatialObject):
...
+            raise ObjectValidationError("Hole chunk ranges must cover the associated table exactly once")
+        expected_offset = offset + count
+    if expected_offset != table_length:
+        raise ObjectValidationError("Hole chunk ranges must cover the associated table exactly once")
+
+
+async def _table_by_hole(
+    table: DownholeDistanceTable | DownholeIntervalTable, data: pd.DataFrame, *, fb: IFeedback
+) -> dict[str, pd.DataFrame]:
+    root = table._context.root_model
```
daniel-kinney-seequent: This looks like it be an instance method based on how it's called and the accessing of `table._context`.

This function adds the assumption that the root_model is a DownholeCollection. That's probably fine. It's worth documenting. Maybe it's worth adding a typed accessor or validation for that. The comment is also relevant for `DownholeCollectionTables.add()`.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -301,3 +472,49 @@ class DownholeCollection(BaseSpatialObject):
...
+    collar_ids = await root.location.hole_id.to_dataframe(fb=fb)
+    categories = collar_ids.iloc[:, 0]
+    result: dict[str, pd.DataFrame] = {}
+    for chunk in (await table.holes.to_dataframe(fb=fb)).itertuples(index=False):
+        hole_id = str(categories.cat.categories[int(chunk.hole_index)])
+        result[hole_id] = data.iloc[int(chunk.offset) : int(chunk.offset) + int(chunk.count)].reset_index(drop=True)
```
daniel-kinney-seequent: It's possible that the same hole_id will show up in different chunks.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -283,9 +363,100 @@ class DistanceTable(SchemaModel):
...
+    @classmethod
+    async def _data_to_schema(cls, data: pd.DataFrame, context: IContext) -> Any:
+        result = await super()._data_to_schema(data, context)
+        unit = data.attrs.get("unit")
+        description = data.attrs.get("attribute_descriptions", {}).get("from")
+        if unit is None and description is not None:
+            unit = description.unit
+        if unit is not None:
+            result["unit"] = unit
+        return result
```
daniel-kinney-seequent: FYI https://github.com/SeequentEvo/evo-python-sdk/pull/316 added a couple helper schema classes for interval tables: https://github.com/SeequentEvo/evo-python-sdk/pull/316. I'm not sure if it's worth consolidating.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -283,9 +363,100 @@ class DistanceTable(SchemaModel):
...
+
+    def __contains__(self, name: object) -> bool:
+        return isinstance(name, str) and self.get(name) is not None
+
+    def names(self) -> list[str]:
+        return [collection.name for collection in self]
+
+    async def add(self, collection: DownholeCollectionEntry, *, replace: bool = False) -> None:
+        existing = self.names()
+        if collection.name in existing and not replace:
```
daniel-kinney-seequent: Can we confirm whether the names must be unique?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -283,9 +363,100 @@ class DistanceTable(SchemaModel):
...
+            raise ObjectValidationError("Collection hole_index is not present in the location holes table")
+        table = collection.distance_table if isinstance(collection, DistanceCollection) else collection.interval_table
+        _validate_chunk_ranges(collection.holes, len(table))
+        schema = await self._data_to_schema([collection], self._obj)
+        if collection.name in existing:
+            self._document[existing.index(collection.name)] = schema[0]
+        else:
+            self._document.append(schema[0])
+
+    def remove(self, *names: str) -> int:
```
daniel-kinney-seequent, line 545: Some notes/questions about this method:
1. Would it be better to raise an exception if something in `names` is missing?
2. If returning an integer is important, would it be better to return the list of names which were removed?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -301,3 +472,49 @@ class DownholeCollection(BaseSpatialObject):
...
+            documents.append(self.location.as_dict())
+        for name in names:
+            collection = self.collections.get(name)
+            if collection is None:
+                raise KeyError(f"Unknown collection '{name}'")
+            documents.append(collection.as_dict())
+        await self.prefetch(data_ids=collect_data_ids(documents), **kwargs)
+
+
+def _validate_chunk_ranges(holes: HoleChunks, table_length: int) -> None:
```
daniel-kinney-seequent: Can this be consolidated with `_validate_hole_chunks()` defined above?

packages/evo-objects/src/evo/objects/utils/downhole.py
```diff
@@ -0,0 +1,67 @@
...
+
+from __future__ import annotations
+
+import numpy as np
+import pandas as pd
+
+__all__ = ["expand_hole_index", "hole_chunks_from_ids"]
+
+
+def hole_chunks_from_ids(hole_ids: pd.Series, *, dtype: pd.CategoricalDtype) -> pd.DataFrame:
```
daniel-kinney-seequent: I'm not sure if this is standard pandas, but it looks awkward to me to take a `pd.CategoricalDtype` argument. Would it be sensible and more clear if it was a more straightforward mapping? Or if you prefer using categories, can the function just accept a category series?

packages/evo-objects/src/evo/objects/utils/downhole.py
```diff
@@ -0,0 +1,67 @@
...
+            end += 1
+        if code in chunks:
+            raise ValueError("Rows for each hole_id must be contiguous")
+        chunks[code] = (start, end - start)
+        start = end
+    return pd.DataFrame(
+        {
+            "hole_index": np.arange(len(dtype.categories), dtype=np.int32),
+            "offset": np.array([chunks.get(code, (0, 0))[0] for code in range(len(dtype.categories))], dtype=np.uint64),
+            "count": np.array([chunks.get(code, (0, 0))[1] for code in range(len(dtype.categories))], dtype=np.uint64),
```
daniel-kinney-seequent: It looks like this code is generating empty chunks for codes that are missing from `hole_ids`. Can I just double check with you that this is desired?

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -117,6 +117,15 @@ def clear_cache(self) -> None:
         """Clear the cache used by this client."""
         self._cache.clear_cache(environment=self._environment, scope=_CACHE_SCOPE)
 
+    def get_static_context(self) -> StaticContext:
```
daniel-kinney-seequent, line 120: Is this used? I didn't spot it when searching.

daniel-kinney-seequent (changes requested):

packages/evo-objects/src/evo/objects/utils/data.py
sergiopeixoto-seequent: No, this is not used directly from the SDK was added to allow others consumers to access the configuration - right now we found at least 2 that are directly accessing the private members like _connector, _cache, and _environmentTo promote reusability and maintain encapsulation, shouldn't this logic be part of a generic SDK instead? So, my suggestion is to keep it here.

packages/evo-objects/src/evo/objects/utils/downhole.py
sergiopeixoto-seequent: Of course, we can use a generic mapping. However, the use of pd.CategoricalDtype was a deliberate choice because hole_index must use the exact, shared categorical codes used by the location and collection tables.

Accepting a generic mapping could hide ordering mismatches, allowing hidden bugs that are very hard to catch.

packages/evo-objects/src/evo/objects/utils/downhole.py
sergiopeixoto-seequent: This is intentional - every category in dtype must produce one chunk, so codes absent from hole_ids receive an explicit zero-count chunk (offset=0, count=0). This preserves complete categorical coverage and is documented and covered by tests.

packages/evo-objects/src/evo/objects/typed/_model.py
sergiopeixoto-seequent: I agree that SchemaUnion could be a useful abstraction in the future, but I don’t think it simplifies this case enough to justify adding it now. DownholeCollectionTables needs different discrimination logic for schema reads (from_to presence) and data serialization (IntervalCollection versus DistanceCollection), so the custom _resolve_item_type and _data_to_schema implementations would still be required. The current SchemaList guard also prevents a union from silently relying on an unset _item_type.

Will wait for @BenLewis-Seequent view.

packages/evo-objects/src/evo/objects/typed/_prefetch.py
sergiopeixoto-seequent: Not sure if you are questioning why 100 was chosen instead of a different value, or if you are questioning whether we can actually have that many tables for a single object.
Regarding the value, the chosen number has no reasoning behind it; it was just a pragmatic decision to set a reasonable threshold.
Regarding the table count, yes, we definitely have DHC  with this many parquet tables. For example, just 4 collections with around 25 attributes each will already reach 100 parquet tables.

packages/evo-objects/src/evo/objects/typed/attributes.py
sergiopeixoto-seequent: Agreed. The schema requires discipline and type, so I restored them as string fields and kept them in the serialized description. unit remains optional because the schema allows it to be omitted for unitless attributes.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: I suggest keeping unit here because it matches the DistanceCollections and IntervalCollections schema: unit is the per-table field under distance/from_to, while distance_unit is the parent DownholeCollection field. depth_unit is not a schema field, so renaming it would introduce a mismatch.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Accept—I agree that table is a cleaner generic name here.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Agreed—the schema does not require zero-based or contiguous lookup keys, I would still like to propose that we enfornce it as part of SDK writing convenction so the documentation now identifies dense categorical codes as an SDK creation convention while reads resolve hole_index through the persisted hole_id lookup keys. is this fine?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Agreed—I replaced the indirect categorical-code lookup with direct [key, value] lookup-table access on reads and an explicit hole-ID-to-key mapping during creation.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Agreed for downloaded objects; this dense-domain validation now applies only while creating a new object, while downloaded objects use their actual persisted lookup keys and therefore support arbitrary integer keys.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: I retained key-set equality for location creation because it states the invariant directly and prevents future changes from allowing a duplicate key to hide a missing hole; it is no longer applied to collection tables.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Agreed for collection tables: they may now contain partial, overlapping, or repeated chunks, and _table_by_hole() concatenates repeated chunks in table-offset order; exact one-time coverage remains only as an SDK creation invariant for the location path to avoid orphaned or ambiguously assigned path rows.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: No reason at all, if you prefer having the caller to make the decision Im now reverting to the previous state.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: I’ve aligned the DHC read wrappers with DataTableAndAttributes.to_dataframe() so callers can optionally select attribute keys while retaining the fb argument.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Since we are ok without not keep backwards compatilibity ive now make sure collection coordinate units come only from DistanceCollection.unit / IntervalCollection.unit, while ordinary attributes retain independently described units.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Sure! Done.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Agreed—there was a small duplication. I’ve reused shared DepthIntervalsTable for collection interval values, while retaining the collection-specific wrapper because it handles attributes and the explicit collection-unit contract. Is this enough or you see more on the consolidation side?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: The schema itself does not enforce unique names, but we uses names as collection-management keys, so we now enforce uniqueness for SDK writes. Legacy objects with duplicates remain readable: lookup warns and returns the first matching collection. are we ok doing so? I find very tricky for consumers having collections with the same name because they will need to inspect the collections to understand if it is the one they want to use.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Returning removed names would help with diagnostics, but remove() may remove multiple legacy documents with the same name, so the integer is the clearest representation of the actual number of documents deleted. The API was intentionally specified as -> int, with unknown names returning 0. Callers that need strict name validation can check membership before calling remove().

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Sure. Done

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Good catch. My bad, Moved to module-level imports.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Sure. Done

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Moved the per-hole grouping logic to a shared instance method and added a documented, validated accessor for the required DownholeCollection root; DownholeCollectionTables.add() now uses the same accessor.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Yes. Collection tables can contain multiple chunks for the same hole_id, since the schema does not enforce one chunk per hole. The read path already groups those chunks by the persisted lookup key, orders them by table offset, and concatenates them. SDK-generated location chunks remain one-per-hole; the one-chunk collection output from our helper is a writer convention, not a read-time assumption.

packages/evo-objects/src/evo/objects/typed/_model.py
BenLewis-Seequent: I don't see a huge difference between both approaches, as it just only really changes where the discrimination logic sits. The advantage of  `SchemaUnion` is it separates it out of the list class, but that isn't a huge problem. This can also be easily changed later without any issues.

packages/evo-objects/src/evo/objects/typed/attributes.py
daniel-kinney-seequent: Is it possible to avoid relaxing `unit` to `Any`?

packages/evo-objects/src/evo/objects/typed/_prefetch.py
daniel-kinney-seequent: At a glance, it just seemed like a big number of concurrent downloads. Usually when I see this sort of thing it tends to be 8ish. There does seem to be some precedent for smaller numberes in `ChunkedIOManager` and `AioTransport`. Would you object to making it something more typical, like 8?

packages/evo-objects/src/evo/objects/utils/data.py
daniel-kinney-seequent: Sure, it makes sense. Thanks for explaining.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: You're right - `unit` matches the schema.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: There might be a nicer way to do it without the special handling in `DownholeCollectionTables`. It would involve specifying and the `DataLocation` field for `unit` for `DistanceTableDistances` and `IntervalTableFromTo`, and altering the parents `DataLocation` scopes. That's just polish though. If it works it's fine.

Oh, one more thing. The readme might need updating. “an explicit collection unit takes precedence over unit metadata on the distance or from column.”

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: Yeah basically. `FromToModel` is also similar to what you did, but it's such a minor thing and not worth reworking.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: Okay, thanks for explaining the rational. It might be worth adding a docstring.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: It looks like it's still there.

packages/evo-objects/src/evo/objects/typed/_model.py
daniel-kinney-seequent: Okay, I still have a minor concern about `cls._item_type` being unset, but at least the use is guarded for now.

packages/evo-objects/src/evo/objects/utils/downhole.py
daniel-kinney-seequent: Do you have a reason why "every category in dtype must produce one chunk"?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: I agree that making it a writing convention makes sense, so it makes perfect sense for helpers like `hole_chunks_from_ids()` to generate `hole_index` in the most sensible way.

But I don't agree that the code in `downhole_collection.py` having to do with `DownholeCollectionData` or `DownholeCollection` should make any assumption about this. So I don't agree with the validation in `DownholeCollectionData._validate_hole_chunks()`, or the dependency on this assumption in `compute_bounding_box()`.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: My main concern is that this prevents `DownholeCollectionData` from being populated with valid schema data, which could easily come from what `DownholeCollection` just fetched. But also, I can imagine valid workflows where holes overlap. I have a harder time coming up with a reason why there would be orphan rows.

My concerns are only theoretical, so comment resolved.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: Resolving this comment - leaving just the one above.

packages/evo-objects/src/evo/objects/utils/downhole.py
daniel-kinney-seequent: Sorry, I’m not sure I understand why a `pd.CategoricalDtype` is the type of the argument.

As I understand it, the dtype defines the categories and their ordering, and pandas derives the integer codes from that ordering. A caller can provide any CategoricalDtype, so they could accidentally use an ordering inconsistent with the rest of the DownholeCollectionData.

I understand that the same hole ID-to-index mapping must be used consistently when generating a DownholeCollectionData. What I’m unclear about is why that mapping should be represented and passed around as a pd.CategoricalDtype, rather than a simpler data structure?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
daniel-kinney-seequent: I agree.

packages/evo-objects/src/evo/objects/typed/_prefetch.py
sergiopeixoto-seequent: Sure. Not a problem. Done in https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/dbbd9985d05efc9bed1fb7fb396aff775540f7c6

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Hope its fixed this time in https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/21d2e6b45ad930ea66cf69f7e51dc56cb9ab7bff

packages/evo-objects/src/evo/objects/typed/attributes.py
sergiopeixoto-seequent: Sure, I made unit a typed as str | Unit | None. 
Be aware that Unit is a local structural protocol for schema-generated enum objects with a string .value, so we retain enum compatibility without adding an evo_schemas dependency.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Hope this is now sorted on https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/1a7a2d609b599bda2628b98a1d252e1107fb2950 Can you confirm?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Should be removed on https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/1a7a2d609b599bda2628b98a1d252e1107fb2950. Can you confirm?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: It should be solved by  https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/1a7a2d609b599bda2628b98a1d252e1107fb2950. Can you confirm?

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: This ones should also be now solved by https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/1a7a2d609b599bda2628b98a1d252e1107fb2950. Please confirm.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: Added on https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/a02ac93b04eacf408c5124ae93aa29dd74d2bda0

packages/evo-objects/src/evo/objects/utils/downhole.py
sergiopeixoto-seequent: Should be now fixed with https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/1a7a2d609b599bda2628b98a1d252e1107fb2950. Please check.

packages/evo-objects/src/evo/objects/utils/downhole.py
sergiopeixoto-seequent: Should now be fixed by https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/1a7a2d609b599bda2628b98a1d252e1107fb2950. Please check.

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
sergiopeixoto-seequent: clarification added to readme on https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/41451256b5b1551a63603987a8e52fbaf410b558

packages/evo-objects/src/evo/objects/typed/attributes.py
daniel-kinney-seequent: Looks good

packages/evo-objects/src/evo/objects/utils/downhole.py
```diff
@@ -0,0 +1,76 @@
...
+import pandas as pd
+
+__all__ = ["expand_hole_index", "hole_chunks_from_ids"]
+
+
+def hole_chunks_from_ids(hole_ids: pd.Series, *, hole_indices: Mapping[str, int] | None = None) -> pd.DataFrame:
+    """Run-length encode contiguous hole IDs into ``[hole_index, offset, count]`` chunks.
+
+    ``hole_indices`` maps hole IDs to their lookup-table keys. When omitted, this
+    creation helper assigns dense zero-based keys in sorted ID order. Persisted
```
daniel-kinney-seequent, line 28: I think this comment is out of date now.

sergiopeixoto-seequent: Updated on https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/ea86f6ebbb2337e20c7502d7358338d8e462e795

packages/evo-objects/src/evo/objects/typed/downhole_collection.py
```diff
@@ -95,26 +108,103 @@ class DownholeCollectionData(BaseSpatialObjectData):
...
-            bboxes.append(self._compute_hole_bounding_box(path_table, collar))
+        collars = self.properties.copy()
+        collars["_hole_index"] = collars["hole_id"].astype(str).map(self.hole_indices)
+        collars_by_index = collars.set_index("_hole_index")
 
+        bboxes = []
+        for hole_index, offset, count in self.holes[["hole_index", "offset", "count"]].itertuples(
+            index=False, name=None
+        ):
+            coordinates = collars_by_index.loc[int(hole_index), _COORDINATE_COLUMNS].to_numpy(dtype=np.float64)
```
daniel-kinney-seequent: This mapping from `hole_index` to `properties` (which is called "collars" in this function) doesn't look like what I was expecting. I thought `hole_index` was a straight lookup into `properties`.

sergiopeixoto-seequent: Fix on https://github.com/SeequentEvo/evo-python-sdk/pull/325/changes/032b3e16eff13aed7740b4e4f65c7bb41e8caed4

daniel-kinney-seequent (approved):

Issue #218, opened 2026-02-23
Bug: Downloading multiple data tables is slow


sergiopeixoto-seequent:
## Bug Description
Retrieving multiple data tables from a Geoscience Object is slow (fetching data is done in sequence).

## Steps to Reproduce
This code relies on the core-compute-task, note that select_and_download_table in the end is calling evo.objects.utils.data.ObjectDataClient.download_table (in the end where the problem happens I think)

`
async def download_parquet_from_data_ids(
    data_ids: list[str],
    geoscience_object: TypedGeoscienceObject,
) -> list[pa.Table]:
    """
    Download parquet files directly from data IDs using parallel HTTP downloads.

    Uses presigned download URLs cached on the TypedGeoscienceObject from its initial
    load() call.

    Args:
        data_ids: List of SHA256 data hashes referencing parquet files.
        geoscience_object: A loaded TypedGeoscienceObject (provides data_urls).

    Returns:
        List of PyArrow tables.

    Raises:
        ValueError: If the geoscience object has not been loaded or has no cached URLs.
    """
    if not data_ids:
        return []

    urls_by_name = geoscience_object.data_urls
    if not urls_by_name:
        raise ValueError(
            "Geoscience object has no cached download URLs. "
            "Ensure load() has been called before downloading parquet data."
        )

    # Resolve download URLs for all data IDs
    download_urls: list[str] = []
    for data_id in data_ids:
        url = urls_by_name.get(data_id)
        if not url:
            available = list(urls_by_name.keys())[:5]
            raise ValueError(
                f"No download URL found for data hash {data_id}. "
                f"Available hashes ({len(urls_by_name)} total): {available}"
            )
        download_urls.append(url)

    # Download all parquet files in parallel
    async def _download_one(session: aiohttp.ClientSession, url: str) -> pa.Table:
        async with session.get(url) as resp:
            resp.raise_for_status()
            return pq.read_table(BytesIO(await resp.read()))

    async with aiohttp.ClientSession() as session:
        tables = await asyncio.gather(*[_download_one(session, url) for url in download_urls])

    return list(tables)

# Collect data IDs and filter out "sig" attributes
    filtered_attributes = []
    data_ids = []
    coroutines = []
    for attribute in collection_attributes:
        filtered_attributes.append(attribute)
        data_ids.append(attribute.values.data)
        attribute_ref_path = (
[...]

PR #331, opened 2026-08-11
Add batch download tables

jamesz3ngsq:
## Description
This PR aims to address the slow download issue raised [here](https://github.com/SeequentEvo/evo-python-sdk/issues/218).

For a Geoscience Object that contains many tables(i.e 100 tables) fetching that data using `download_table` is very slow as it will make a metadata request and fetch the object definition each time for every table to prepare the context for downloading. To address this, I've introduced a new method `download_tables` this will make a single metadata request to prepare the context for downloading, and the specified tables will be downloaded concurrently.

## Checklist

- [x ] I have read the contributing guide and the code of conduct


packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -46,6 +47,7 @@
 logger = logging.getLogger("object.data")
 
 _DATA_ID_KEY = "data"  # The key used to identify data references in geoscience objects.
+_DEFAULT_MAX_CONCURRENCY = 4
```
tim-hamblin-seequent: How did you choose this number?

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -287,20 +289,109 @@ async def download_table(
...
+
+        :return: A mapping of data IDs to prepared download contexts.
+
+        :raises DataNotFoundError: If a data ID is not associated with this object version.
+        """
         # Import here to avoid circular import.
         from ..client import ObjectAPIClient
-        from ..parquet import ParquetDownloader
+
+        data_ids = list(dict.fromkeys(str(data_id) for data_id in data_identifiers))
```
tim-hamblin-seequent: Is this just trying to convert the `data_identifiers` to strings? You shouldn't need to pass it through a dictionary, a list comprehension would be simplest: `[str(data_id) for data_id in data_identifiers]`

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -287,20 +289,109 @@ async def download_table(
...
+        if not data_ids:
```
tim-hamblin-seequent: I prefer to be explicit with these kinds of checks (not sure what the standard is in this codebase though). Also if you do this at the very start of the function you can avoid need to do the import and the list comprehension (just check `data_identifiers`)

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -287,20 +289,109 @@ async def download_table(
...
+
+        :raises ValueError: If ``max_concurrency`` is less than one.
+        :raises DataNotFoundError: If a data ID is not associated with this object version.
+        :raises TableFormatError: If a table does not match its expected format.
+        :raises SchemaValidationError: If a table has a different number of rows than expected.
+        """
+        if max_concurrency < 1:
+            raise ValueError("max_concurrency must be at least 1")
+
+        table_infos_by_data_id: dict[str, dict] = {}
```
tim-hamblin-seequent: This could be a dictionary comprehension, I'm not sure you need to use `setdefault` do you?
`{str(table_info['data']): table_info for table info in table_infos_by_data_id}`

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -287,20 +289,109 @@ async def download_table(
...
+        for table_info in table_infos:
+            table_infos_by_data_id.setdefault(str(table_info["data"]), table_info)
+        if not table_infos_by_data_id:
```
tim-hamblin-seequent: Is this even a valid usage of this function? Should it just raise an error?

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -287,20 +289,109 @@ async def download_table(
...
+        feedbacks = split_feedback(fb, [1.0] * len(data_ids))
+        semaphore = asyncio.Semaphore(max_concurrency)
+
+        async def download_one(data_id: str, table_info: dict, table_fb: IFeedback) -> tuple[str, pa.Table]:
+            async with semaphore:
+                table = await self._download_prepared_table(downloads[data_id], table_info, table_fb)
+            return data_id, table
+
+        tasks = [
+            asyncio.create_task(download_one(data_id, table_info, table_fb))
```
tim-hamblin-seequent: I think you don't need to call `asyncio.create_task` here as you are calling `gather` straight away which will create those tasks for you, you can just create a list of the raw coroutines that are returned from `download_one(data_id, table_info, table_fb)`. I think there is slightly less overhead in using the raw coroutines in `gather`

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -287,20 +289,109 @@ async def download_table(
...
+            for (data_id, table_info), table_fb in zip(table_infos_by_data_id.items(), feedbacks, strict=True)
```
tim-hamblin-seequent: `strict=True` seems unnecessary here, will the two lengths ever be different?

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -287,20 +289,109 @@ async def download_table(
...
+        ]
+        try:
+            results = await asyncio.gather(*tasks)
+        except BaseException:
+            for task in tasks:
```
tim-hamblin-seequent: Ah I see that you are actually accessing the tasks here, so maybe my earlier comment is not true!

Generally we try to avoid catching generic exceptions. What is actually going on in this except clause? You cancel the tasks then re-gather them but raise the exceptions? What's going on here?

tim-hamblin-seequent (commented):
A few comments. I think this make sense to me after having a dig into it: you are leveraging the fact that `prepare_data_download` offers the ability to specify multiple `data_identifiers` and you are exposing it via `download_tables`. This seems all good to me, but it will need the check off from a maintainer.

packages/evo-objects/src/evo/objects/utils/data.py
jamesz3ngsq: The gather function doesn't cancel other coroutines if one of them fails, so here I'm trying to cancel each task, ensure they are all cancelled with the gather then return the exception

packages/evo-objects/src/evo/objects/utils/data.py
jamesz3ngsq: I was using setdefault here to retain the first table_info for a given data_id you receive, in case of duplicates, but probably better to use a comprehension and caller can ensure no duplicates

packages/evo-objects/src/evo/objects/utils/data.py
jamesz3ngsq: No reason in particular besides 4 feeling like a safe number 😂

How do you go about deciding the default for things like this?

packages/evo-objects/src/evo/objects/utils/data.py
tim-hamblin-seequent: Comprehension will take the last `table_info` so if you really need it to be the first `table_info` you could probably just iterate over it backwards, but I don't see why it would matter which one you take if they are duplicates.

packages/evo-objects/src/evo/objects/utils/data.py
tim-hamblin-seequent: Good question. This is always tricky as it will depend on how semaphore works under the hood, the hardware that it's running on, the internet speed, the moon phase etc. So given lots of different users are using it then it can be really tricky to find a sensible value.

I would be interested in some benchmarking. I'm curious how the efficiency gain changes as this number changes. My guess as it will improve as the number increases to a point then either stop/decrease after that point because of some overhead in checking on each operation that we are waiting for.

Or maybe it won't change at all, and we can just use `1` 🤷 You never know.

packages/evo-objects/src/evo/objects/utils/data.py
tim-hamblin-seequent: Ah yeah async programming is tricky. I had a quick read and maybe: `asyncio.TaskGroup` could be of use as it sounds like it has the behaviour you want?

packages/evo-objects/src/evo/objects/utils/data.py
jamesz3ngsq: Did some really rough benchmarking, testing the speed between using 1, 2, 4, 8, 16 for the concurrency count. Does seem to be a decent performance improvement going from 1 to 4 but beyond that not so much so will be keeping it at 4.

peter-kipping-seequent: Can we add a comment to this code saying what your last comment says. Something like
"Decent performance improvement going from 1 to 4 but beyond that not so much"

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -287,20 +289,102 @@ async def download_table(
...
+
+        :raises DataNotFoundError: If a data ID is not associated with this object version.
+        """
         # Import here to avoid circular import.
         from ..client import ObjectAPIClient
-        from ..parquet import ParquetDownloader
+
+        data_ids = list({str(data_id) for data_id in data_identifiers})
+
+        if len(data_ids) == 0:
```
peter-kipping-seequent: Theoretically you could move this check above the previous line to be 'more optimal' since your adding in an early exit condition. So something like 
`
if len(data_identifiers) == 0:
`

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -287,20 +289,102 @@ async def download_table(
         :raises TableFormatError: If the data does not match the expected format.
         :raises SchemaValidationError: If the data has a different number of rows than expected.
         """
+        data_id = str(table_info["data"])
+        downloads = await self._prepare_data_downloads(object_id, version_id, [data_id])
+        return await self._download_prepared_table(downloads[data_id], table_info, fb)
+
+    async def _prepare_data_downloads(
```
peter-kipping-seequent, line 296: Im not the biggest fan on this function name, especially since its just shadowing the actual 'prepare_data_downloads' on a client.
However i cant really think of a better one right now.
If someone has a better suggestion im all ears, but what does 'prepare data downloads' mean?
Isnt it more like 'fetch list of downloads" or something?

packages/evo-objects/src/evo/objects/utils/data.py
```diff
@@ -287,20 +289,102 @@ async def download_table(
...
+    async def _download_prepared_table(self, download: ObjectDataDownload, table_info: dict, fb: IFeedback) -> pa.Table:
+        """Download and load one table using a prepared download context.
+
+        :param download: The prepared context for the table's data.
+        :param table_info: The table info that defines the expected format.
+        :param fb: A feedback object for tracking download progress.
+
+        :return: A pyarrow table loaded directly from the parquet file.
+        """
+        from ..parquet import ParquetDownloader
```
peter-kipping-seequent, line 331: So the line above that has imports says it does it to avoid a circular reference?
Is this being imported here for the same reason? 
Then it should have the comment again.

Out of scope of this ticket to fix the core issue i suppose but i did want to note my general objection to that being the solution here :(

peter-kipping-seequent (changes requested):
Adding some comments to the PR, however the main issue is actually that this approach isnt viable in 3.10, which we formally support on this SDK.
Taskground was introduced in 3.11, do you'll have to change the implementation of this to a 3.10 compatible concept.

jamesz3ngsq:
Hey Peter, I've implemented your suggestions and used an async method compatible with 3.10. Cheers.

peter-kipping-seequent (approved):
Nice, 3.10 compatible and a bit cleaner. nice work @jamesz3ngsq.

GriffinBaxterSeequent (approved):

tim-hamblin-seequent (approved):
LGTM!

GriffinBaxterSeequent (approved):

PR #333, opened 2026-08-18
Handle some race conditions in downloading data

grantroch:
## Description

This PR attempts to fix a couple race conditions that can happen when downloading data.

1. In `download_file` it can be the case that two different processes download the same file at the same time. They will both download different temporary files, but if one finishes downloading and renames before the other one, the second process will throw a `PermissionError`. If the file exists, the `PermissionError` is ignored and the destination path points to a valid file.
2. A lock was added to prevent concurrent coroutines in the same process from duplicating a download. This is rarer as the user isn't likely to be downloading the same exact file from two coroutines. Subsequent callers will wait and use the cached result.

## Checklist

### DIFF
--- a/packages/evo-sdk-common/pyproject.toml
+++ b/packages/evo-sdk-common/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "evo-sdk-common"
 description = "Python package that establishes a common framework for use by client libraries that interact with Seequent Evo APIs"
-version = "0.5.26"
+version = "0.5.27"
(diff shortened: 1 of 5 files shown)


GriffinBaxterSeequent (approved):
Nice one! Do note that you'll want to bump the version for evo-sdk-common

AdamFreemanSeequent (approved):

GriffinBaxterSeequent (approved):

PR #341, opened 2026-08-27
Blockmodel support additional units

### LINKED ISSUE #320: Bug: Cannot Import Block Model Units
## Bug Description
`evo.blockmodels.typed` fails to import because its init.py imports from evo.common.typed - which does not exist.

## Steps to Reproduce
import anything from [units.py](packages/evo-blockmodels/src/evo/blockmodels/typed/units.py). 

## Expected Behavior
Imports as expected

## Actual Behavior
Traceback:
`from evo.common.typed import BoundingBox, Point3, Size3d, Size3i
ModuleNotFoundError: No module named 'evo.common.typed'`

jamie-spyker-sqnt:
## Description

- Extended blockmodel unit support with dimensionless, category, and angle unit types.
- Updated type/unit ordering to match expected user-facing order
- Updated units to be an enum type. using `str enum` to maintain compatibility with existing SDK logic
- Removed `VOLUME_BY_VOLUME` unit type, no known reference to this outside the SDK
- Fix invalid imports in package `__init__` file, resolves: https://github.com/SeequentEvo/evo-python-sdk/issues/320
- Resolve to unknown if unit does not exist in the SDK

## Checklist


packages/evo-blockmodels/src/evo/blockmodels/typed/units.py
```diff
@@ -34,14 +34,16 @@
 class UnitType(Enum):
     """Types of units supported by the Block Model Service."""
 
-    LENGTH = "LENGTH"
+    MASS_PER_MASS = "MASS_PER_MASS"
+    MASS_PER_VOLUME = "MASS_PER_VOLUME"
+    DIMENSIONLESS = "DIMENSIONLESS"
```
AndreLobatoSeequent: Inside BMS we classify the `DIMENSIONLESS` as `GENERIC_DIMENSIONSLESS`. I suppose the reasoning is to differentiate from the other dimensionless units such as MASS_PER_MASS, etc..

packages/evo-blockmodels/src/evo/blockmodels/typed/units.py
```diff
@@ -86,60 +88,73 @@ class Units:
...
     SHORT_TON_PER_CUBIC_FOOT = "ton[US]/ft3"
+    TONNES_PER_CUBIC_METRE = "t/m3"
 
-    # Value units
-    DOLLARS_PER_TONNE = "$/t"
-    DOLLARS_PER_SHORT_TON = "$/ton[US]"
-    DOLLARS = "$"
+    # Dimensionless units
+    COUNT = "count"
+    PERCENTAGE = "%"
```
AndreLobatoSeequent: We are looking into updating this to PERCENT, also update the current PERCENT to PERCENT_MASS. So maybe we could wait on this before merging?

packages/evo-blockmodels/src/evo/blockmodels/typed/units.py
jamie-spyker-sqnt: Would you see value in following the same naming conventions here? Right now all of these named types are kept quite simple and feel clear to me. `MASS_PER_MASS` feel differentiated as it is inherently more specific in its name.

packages/evo-blockmodels/src/evo/blockmodels/typed/units.py
jamie-spyker-sqnt: as was discussed elsewhere; looks like we will be dropping this for now 👍

packages/evo-blockmodels/src/evo/blockmodels/typed/units.py
jamie-spyker-sqnt: Looks like we need to update the enum value to be "GENERIC_DIMENSIONLESS" to match BMS otherwise am running into errors.

rohancrookbain-seequent (approved):

AndreLobatoSeequent (approved):

PR #345, opened 2026-08-31
Add column groups support

AndreLobatoSeequent:
## Description

Adds full column-group support to the high-level `BlockModelAPIClient` for the block model service. Previously the SDK could only read groups off a version; callers had to drop down to the generated endpoint layer to create, assign, or modify them. This change exposes the complete group lifecycle — create, assign, read, update and delete — through the public client and data model.

Column groups are a preview feature: the client must be constructed with `preview=True` to use them, consistent with column tags. (**This is no longer required, this will be merged after the feature is out of API-Preview.**)

The design goal is for the SDK to **mirror the service**: where the service addresses columns by their qualified (`group▸title`) title, the caller supplies that same qualified title. The SDK does not invent a plain-title abstraction, rename your data, or perform hidden version lookups — this avoids ambiguity when the same title is reused across groups.

**Write path**

- Create / update / delete groups — new `BlockModelAPIClient.update_groups(bm_id, new=, update=, delete=, comment=)`. Create groups with `GroupDefinition` (title, parent group, missing-column policy, tags, hidden flag); rename, re-parent, change policy, replace tags or toggle visibility with `GroupMetadataUpdate`; delete by title. Any combination can be applied in a single call.
- Group a **new** column — the `column_groups` parameter on `add_new_columns`, `add_new_subblocked_columns`, `update_block_model_columns` and `update_subblocked_columns` maps a new column's qualified title (its key in `data`, e.g. `Assays▸Cu`) → the group it belongs to (`Assays`). The SDK derives the title by stripping the group prefix.
- Move or ungroup an **existing** column — `update_column_metadata` with `ColumnMetadataUpdate(group=...)`. A column's group is metadata, so moving it needs no data upload. Address the column by its **current** qualified title and set `group` to the new group's qualified title (or `""` to ungroup).

**Custom qualified-title separator**

The character that joins a group's title to a column's title in a qualified title defaults to `▸` (`QUALIFIED_TITLE_SEPARATOR`). A block model may use a different single-character separator, so both the write and read paths accept an optional `separator` argument:

- Write path — `add_new_columns`, `add_new_subblocked_columns`, `update_block_model_columns`, `update_subblocked_columns`.
- Read path — `query_block_model_to_cache`, `query_block_model_as_table`.

On every one of these, `separator` is used to parse the qualified titles the caller supplies and is forwarded to the service (as `qualified_title_separator`) for that request; on the read path it also governs how returned qualified headers are rendered. The default is never sent on the wire, so existing behaviour is unchanged; the argument is only serialised when a non-default separator is provided. This matters because a model written with a non-default separator (whose stored titles may legitimately contain `▸`) can only be queried back by declaring the same separator — otherwise the service rejects the request.

**Column titles**

The caller keys `data` by each column's title: a plain title for an ungrouped column, or the qualified `group▸title` for a grouped one. The SDK uploads the table verbatim and never renames columns. Two opt-in helpers are provided to build these titles from plain-titled data:

- `get_qualified_title(group, title)` — builds a single qualified title.
- `qualify_column_titles(data, {title: group})` — renames a plain-titled table and returns `(renamed_table, column_groups)` ready to pass straight to the client.

**Read path**

- `Version` / `ListingVersion` gain bridge helpers so a caller who wrote a group by title can find it again on the returned version without hand-rolling a lookup: `group_by_uuid`, `group_for_column`, `qualified_group_title` and `group_by_qualified_title`.

**Public API**

- New exports from `evo.blockmodels.data`:
  - `GroupDefinition`
  - `GroupMetadataUpdate`
  - `MissingColumnPolicy`
  - `QUALIFIED_TITLE_SEPARATOR`
  - `get_qualified_title` and `qualify_column_titles`
- `get_qualified_title`, `qualify_column_titles` and `QUALIFIED_TITLE_SEPARATOR` are also re-exported from the `evo.blockmodels` package root.

#### Notes

- New parameters are appended to existing method signatures to preserve positional-argument compatibility.
- Existing columns (in `update_columns` / `delete_columns`) are addressed by the exact title the service currently stores them under: the qualified `group▸title` if grouped, or the plain title if not. There is no hidden lookup — the caller states the current reference.
- Moving or ungrouping an existing column is a metadata-only operation (no data upload) via `update_column_metadata`; `ColumnMetadataUpdate` accepts a `group` field (`""` ungroups). The `column_groups` parameter on the column methods only groups **new** columns.
- Groups cannot be referenced during `create_block_model` (none exist yet); the documented flow is create the model → `update_groups(new=...)` → add columns with `column_groups`.
- Input models (`GroupDefinition`, `GroupMetadataUpdate`, `ColumnMetadataUpdate`) reject unknown fields to catch typos early.
- Adds test coverage for grouping new columns, metadata-only moving/ungrouping of existing columns, data-only updates of grouped columns, group create/update/delete, the `get_qualified_title` / `qualify_column_titles` helpers, per-request `group_missing_column_override`, the custom `separator` (parsing qualified titles and forwarding `qualified_title_separator`, default omitted), preview-header propagation, input validation, and all bridge helpers.
- The qualified-title separator is per-request: `QUALIFIED_TITLE_SEPARATOR` (`▸`) is only the default. Callers on a non-default separator pass it explicitly via `separator` on both the write and read methods; the constant is never hard-coded internally, so those callers are not broken. The generated `UpdateDataLite` and `QueryCriteria` endpoint models carry the `qualified_title_separator` field.

## Acceptance criteria

- [x] Column groups can be created, assigned, read, updated and deleted through the public `BlockModelAPIClient`.
- [x] New columns can be grouped on add/update; existing columns can be moved/ungrouped via metadata only (no data upload).
- [x] A non-default qualified-title separator can be supplied per request on both the write and read paths, is used to parse qualified titles, and is forwarded to the service as `qualified_title_separator`; the default `▸` is never sent on the wire. A model written with a non-default separator can be queried back by declaring the same separator.
- [x] Public helpers and exports (`get_qualified_title`, `qualify_column_titles`, `QUALIFIED_TITLE_SEPARATOR`) are available from `evo.blockmodels.data` and the package root.
- [x] Test coverage accompanies each of the above.

## Checklist


packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -709,6 +759,9 @@ async def _add_new_columns(
         :param geometry_change: Whether the geometry of the block model is changing.
         :param tags: A dictionary mapping column names within `data` to their tags object. Column tags are a preview
             feature; the client must be constructed with ``preview=True`` to use them.
+        :param column_groups: A dictionary mapping column names within `data` to the qualified title of the group the
+            column should be placed in. Column groups are a preview feature; the client must be constructed with
+            ``preview=True`` to use them.
```
AndreLobatoSeequent: > Column groups are a preview feature; the client must be constructed with
            ``preview=True`` to use them

We are aiming to remove api-preview flag very soon, perhaps we could avoid this sentence.

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -1003,6 +1092,51 @@ def _to_values(value: str | None | ColumnMetadataUpdate) -> models.UpdateMetadat
...
+        *,
+        new: list[GroupDefinition] | None = None,
+        update: dict[str, GroupMetadataUpdate] | None = None,
+        delete: list[str] | None = None,
+        comment: str | None = None,
+    ) -> Version:
+        """Create, update, and/or delete column groups on a block model.
+
+        This method manages group definitions without requiring data upload or cache configuration. Any
+        combination of ``new``, ``update`` and ``delete`` can be supplied in a single call.
```
AndreLobatoSeequent, line 1215: I'm a bit unsure about this. Is it possible to do groups update without data update?

AndreLobatoSeequent: No it's not possible, so removed that bit. Moved into update_columns methods

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -136,6 +141,66 @@ def _version_listing_from_model(version: models.ListingVersion) -> ListingVersio
...
+        new=[_group_lite_from_definition(definition) for definition in (new or [])],
+        update_metadata=[
+            models.GroupUpdateMetadataLite(title=title, values=_group_values_from_update(values))
+            for title, values in (update or {}).items()
+        ],
+        delete=list(delete or []),
+    )
+
+
+def _qualified_heading(group: str | None, title: str) -> str:
```
rohancrookbain-seequent:
```suggestion
def _qualified_title(group: str | None, title: str) -> str:
```

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -136,6 +144,57 @@ def _version_listing_from_model(version: models.ListingVersion) -> ListingVersio
...
+    """
+    if not group:
+        return heading
+    prefix = f"{group}{_QUALIFIED_TITLE_SEPARATOR}"
+    if not heading.startswith(prefix):
+        raise MissingColumnInTable(
+            f"column '{heading}' is declared in group '{group}' but its data heading is not the qualified "
+            f"title '{prefix}<leaf>'. Key the data by its exact upload heading (see qualify_column_titles)."
+        )
+    return heading[len(prefix) :]
```
rohancrookbain-seequent: **issue:** I dislike that this uses the terms "leaf title" and "heading", which are not prevalent in the documentation; it would be more helpful if the terms "bare title" and "fully qualified title" were used.
Additionally I am not sure what `<leaf>` is doing in the error message.

Also, **suggestion:** `heading.removeprefix(prefix)` can be used instead of slicing

rohancrookbain-seequent (commented):
Partially reviewed; will have a more complete look soon!

packages/evo-blockmodels/src/evo/blockmodels/client.py
AndreLobatoSeequent: Updated the wording all around. also abide by suggestion, good catch, much cleaner that way.

packages/evo-blockmodels/src/evo/blockmodels/data.py
```diff
@@ -14,43 +14,111 @@
...
     "ListingGroup",
     "ListingVersion",
+    "MissingColumnPolicy",
     "OctreeGridDefinition",
     "RegularGridDefinition",
     "ResolvedGroup",
     "Version",
+    "get_qualified_title",
+    "qualify_column_titles",
 ]
```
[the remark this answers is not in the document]
AndreLobatoSeequent: Updated PR description with updated function names.

packages/evo-blockmodels/src/evo/blockmodels/data.py
```diff
@@ -279,6 +404,68 @@ def __repr__(self) -> str:
...
+        :param column: A column from this version, or a column title.
+        :return: The column's group, or ``None`` if the column is ungrouped or unknown.
+        """
+        if isinstance(column, str):
+            column = next((c for c in self.columns if c.title == column), None)
+            if column is None:
+                return None
+        if column.group_uuid is None:
+            return None
+        return self.group_by_uuid(column.group_uuid)
```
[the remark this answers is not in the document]
AndreLobatoSeequent: Updated to deal with the potential ambiguity. Only accepts a Column/ListingColumn object instead.

packages/evo-blockmodels/tests/test_group_helpers.py
```diff
@@ -0,0 +1,249 @@
...
+            parent_group_uuid=uuid.uuid4(),
+            missing_column_policy=MissingColumnPolicy.INHERIT,
+            resolved_missing_column_policy=MissingColumnPolicy.USE_PREVIOUS,
+        )
+        object.__setattr__(self.version, "groups", [*self.version.groups, orphan])
+        self.assertEqual(self.version.qualified_group_title(orphan), "Orphan")
+
+
+if __name__ == "__main__":
+    unittest.main()
```
rohancrookbain-seequent: **nitpick:** Shouldn't this be at the end of the file?

rohancrookbain-seequent (approved):
Looks good :)

packages/evo-blockmodels/tests/test_group_helpers.py
AndreLobatoSeequent: Shouldn't be there at all. Removing...

packages/evo-blockmodels/src/evo/blockmodels/client.py
AndreLobatoSeequent: Comment removed

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -722,20 +805,28 @@ async def _add_new_columns(
...
         if unknown_unit_columns:
             raise MissingColumnInTable(f"units reference columns that are not being added: {unknown_unit_columns}")
         unknown_tag_columns = set(tags) - new_column_names
         if unknown_tag_columns:
             raise MissingColumnInTable(f"tags reference columns that are not being added: {unknown_tag_columns}")
+        unknown_group_columns = set(column_groups) - new_column_names
+        if unknown_group_columns:
+            raise MissingColumnInTable(
+                f"column_groups reference columns that are not being added: {unknown_group_columns}"
+            )
```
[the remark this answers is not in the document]
AndreLobatoSeequent: No longer requires API-preview

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -1003,6 +1155,51 @@ def _to_values(value: str | None | ColumnMetadataUpdate) -> models.UpdateMetadat
...
+        :param new: Definitions of new groups to create.
+        :param update: A dictionary mapping the qualified title of an existing group to the metadata update
+            to apply to it. Use :class:`GroupMetadataUpdate` to rename, re-parent, change the missing-column
+            policy, replace tags, or toggle the hidden flag.
+        :param delete: Qualified titles of groups to delete.
+        :param comment: An optional comment describing the changes. This is max 250 characters.
+        :return: The new version of the block model with the updated groups.
+        """
+        if not new and not update and not delete:
+            raise ValueError("At least one of 'new', 'update' or 'delete' must be provided.")
```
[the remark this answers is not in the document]
AndreLobatoSeequent: This is no longer required, no need to enforce.

packages/evo-blockmodels/tests/test_group_helpers.py
```diff
@@ -0,0 +1,249 @@
...
+
```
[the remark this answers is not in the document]
AndreLobatoSeequent: Removed

packages/evo-blockmodels/tests/test_update.py
```diff
@@ -891,6 +898,75 @@ async def test_update_column_metadata_with_preview_sends_header(self) -> None:
...
+
+    async def test_update_column_metadata_ungroups_column(self) -> None:
+        """An existing grouped column is ungrouped as a metadata-only update by sending ``group=""``."""
+        self.transport.set_request_handler(
+            UpdateRequestHandler(
+                update_result=UPDATE_RESULT,
+                job_response=JobResponse(job_status=JobStatus.COMPLETE, payload=UPDATED_VERSION),
+            )
+        )
+        await self.bms_client_without_cache.update_column_metadata(
```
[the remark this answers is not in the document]
AndreLobatoSeequent: Removed from API-preview

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -968,16 +1114,22 @@ async def update_column_metadata(
...
+        - A :class:`ColumnMetadataUpdate` sets any combination of unit ID, tags and/or group. Only the
           fields explicitly set on the object are sent; unset fields are left untouched. Set
-          ``tags={}`` to clear a column's tags, or ``unit_id=None`` to clear its unit.
+          ``tags={}`` to clear a column's tags, ``unit_id=None`` to clear its unit, or ``group=""`` to
+          move the column out of any group.
+
+        A column's group is metadata, so it can be moved (or ungrouped) here without re-uploading its
+        data. Address the column by the title the service currently stores it under: its qualified title
+        (``group▸title``) if it is currently grouped, or its plain title if it is not. Set
+        ``ColumnMetadataUpdate(group=...)`` to the target group's qualified title (or ``""`` to ungroup).
```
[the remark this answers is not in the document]
AndreLobatoSeequent: API-Preview is no longer required for this PR.

packages/evo-blockmodels/src/evo/blockmodels/client.py
```diff
@@ -136,6 +141,58 @@ def _version_listing_from_model(version: models.ListingVersion) -> ListingVersio
...
+    """
+    if not group:
+        return column_title
+    prefix = f"{group}{_QUALIFIED_TITLE_SEPARATOR}"
+    if not column_title.startswith(prefix):
+        raise MissingColumnInTable(
+            f"column '{column_title}' is declared in group '{group}' but its column title is not the qualified "
+            f"title '{prefix}<title>'. Key the data by each column's exact title (see qualify_column_titles)."
+        )
+    return column_title.removeprefix(prefix)
```
mikecowie-seequent: **issue** - the API supports providing a different group separator. That's not being reflected down here, meaning anyone opting into that (likely rare) path, breaks their ability to use the SDK (at least insofar as their call gets to this helper or any others that reference this constant).

I would suggest that the constant `_QUALIFIED_TITLE_SEPARATOR` should not be used anywhere except the public interface of the SDK functions - and in that place only as a default, likely implicitly from the data model's field default. Instead, the qualified title separator should always be getting passed explicitely into where it's used.

mikecowie-seequent (changes requested):
One meaningful spot

packages/evo-blockmodels/src/evo/blockmodels/client.py
AndreLobatoSeequent: Good point. I will add the support for defining the separator.

AndreLobatoSeequent: Ok implemented now, in both update and query methods.

mikecowie-seequent (approved):

PR #348, opened 2026-09-09
Add kriging diagnostic outputs to the task SDK client

amirabbas-jalali-bentley:
## Description

Adds support for the kriging task's diagnostic outputs to the `evo-compute` SDK client.

The kriging task can write per-location diagnostics onto the target object alongside the estimate — kriging variance, slope of regression, sample counts, search distances, and so on. These live under `target.diagnostics` in the task schema, and were previously not reachable through the SDK.

### Parameters

`KrigingParameters` gains a `diagnostics` field taking a `KrigingDiagnostics`. It is folded into `target.diagnostics` by the existing model serializer, the same way `source_filter` and `target_filter` are already folded into `source.filter` / `target.filter` — so callers keep passing a typed attribute (or a `Target`) for `target` and do not have to hand-build the nested object.

Every diagnostic accepts:

- `True` — create an attribute using the recommended name (`RECOMMENDED_DIAGNOSTIC_NAMES`, e.g. `KV`, `SoR`, `NS`)
- a string — create an attribute with that name
- a typed attribute from the target object — update it if it exists, otherwise create it
- an explicit `CreateAttribute` / `UpdateAttribute`

```python
params = KrigingParameters(
    source=pointset.attributes["CU_pct"],
    target=block_model.attributes["CU_estimate"],
    variogram=variogram,
    search=SearchNeighborhood(ellipsoid=search_ellipsoid, max_samples=16),
    diagnostics=KrigingDiagnostics(
        kriging_variance=True,        # -> "KV"
        slope_of_regression=True,     # -> "SoR"
        num_samples="sample_count",   # -> "sample_count"
    ),
)
```

The recommended names match the conventions used elsewhere in the product so results are familiar once imported. They are the schema's suggested names rather than server-side defaults, so the SDK always sends an explicit attribute name.

### Validation

Three mistakes are caught before the payload is built, so they surface as a `ValidationError` on `KrigingParameters` rather than a failed or partially applied job:

- **Every output needs its own attribute.** Outputs are compared by attribute name — including the name of a typed attribute being updated, which its reference does not carry — and updates are also compared by reference. This covers a diagnostic colliding with `target.attribute`. A hand-written `UpdateAttribute` expression carries no name, so it is only compared with other references.
- **Diagnostics belong to the target object.** A typed attribute is reduced to a name or reference, which is then resolved against `target.object` — so an attribute taken from a different object would silently resolve to nothing, or to the wrong attribute. The originating object is remembered during validation and compared with the target, ignoring any `?version=` suffix. Attributes that carry no parent object cannot be checked and are left alone.
- **An existing attribute must be able to hold the diagnostic's values.** As in the task schema, updating an existing attribute needs a boolean attribute for `valid`, an integer one for the `num_*` counts, and a floating-point one for the rest. A newly created attribute takes the diagnostic's type, so it is not checked.

### Results

The task returns every diagnostic key, with `null` for the ones that were not requested. New `KrigingDiagnosticsResult` and `KrigingTargetResult` models parse that, and `KrigingResult.diagnostics` exposes only the diagnostics that were actually written, keyed by diagnostic name:

```python
for diagnostic, attribute in result.diagnostics.items():
    print(f"{diagnostic}: {attribute.name}")
```

They are also listed in the result summary. Because diagnostics are written onto the target object, `result.to_dataframe()` returns them alongside the estimate.

### Other changes

- `attribute_spec()` factored out of `_validate_target_attribute` in `tasks/common/source_target.py` and shared, so a diagnostic accepts the same typed attributes as a target does.
- `KrigingDiagnostics` exported from `evo.compute.tasks` and `evo.compute.tasks.geostatistics`.
- `tests/test_tasks.py` fixtures build a `KrigingTargetResult` instead of a bare `TaskTarget`, following the retyped `KrigingResultModel.target`.

### Backwards compatibility

No breaking change to the SDK's public surface — `diagnostics` is optional and defaults to `None`, so existing kriging calls produce an identical payload.

## Testing

- `uv run --package evo-compute pytest packages/evo-compute/tests` — 446 passed (33 new tests, plus 104 subtests, covering the accepted shorthands, the serialized payload, the parsed result, and the three validation rules).
- `ruff check` and `ruff format --check` clean across the repository.
- The generated payload and a realistic result payload were validated against the kriging task's JSON schema with `jsonschema` (Draft 2020-12) — no errors on either side.
- Notebooks were verified to parse; they need live Evo credentials to execute, so they were not run end to end.

## Documentation

- `packages/evo-compute/docs/examples/kriging.ipynb` — new "Request Diagnostics" section.
- `code-samples/geoscience-objects/running-kriging-compute/` — the walkthrough now requests diagnostics, prints the attributes that were written, and includes them in the results DataFrame; the markdown documents the shorthands and lists all 16 diagnostics with their recommended names.

The generated API reference under `mkdocs/` is left to the documentation workflow.

## Checklist

### DIFF
--- a/mkdocs/site/packages/evo-compute/typed-objects/source-target/UpdateAttribute.html
+++ b/mkdocs/site/packages/evo-compute/typed-objects/source-target/UpdateAttribute.html
@@ -39,7 +39,7 @@
                                 </a>
                             </li>
                             <li class="nav-item">
-                                <a rel="next" href="../../../evo-files/index.html" class="nav-link">
+                                <a rel="next" href="attribute_spec.html" class="nav-link">
                                     Next <i class="fa fa-arrow-right"></i>
                                 </a>
                             </li>
--- a/packages/evo-compute/src/evo/compute/tasks/geostatistics/kriging.py
+++ b/packages/evo-compute/src/evo/compute/tasks/geostatistics/kriging.py
@@ -40,28 +40,46 @@ import pandas as pd
 from evo.common import IContext, IFeedback
 from evo.objects import ObjectSchema
 from evo.objects.typed import BaseObject, object_from_reference
-from pydantic import BaseModel, Field, SerializerFunctionWrapHandler, model_serializer
+from pydantic import (
+    BaseModel,
+    Field,
+    PrivateAttr,
+    SerializerFunctionWrapHandler,
+    ValidationInfo,
+    ValidatorFunctionWrapHandler,
+    field_validator,
+    model_serializer,
+    model_validator,
+)
 
 # Import shared components
 from ..common import (
     AnySourceAttribute,
     AnyTargetAttribute,
+    CreateAttribute,
     Filter,
     GeoscienceObjectReference,
     SearchNeighborhood,
+    UpdateAttribute,
+    attribute_spec,
 )
-from ..common.results import TaskTarget
+from ..common.results import TaskAttribute, TaskTarget
 from ..common.runner import TaskRunner
+from ..common.source_target import AnyTypedAttribute
 
 __all__ = [
     # Kriging-specific (users import from evo.compute.tasks.kriging)
+    "RECOMMENDED_DIAGNOSTIC_NAMES",
     "BlockDiscretisation",
     "Filter",
+    "KrigingDiagnostics",
+    "KrigingDiagnosticsResult",
     "KrigingMethod",
     "KrigingParameters",
     "KrigingResult",
     "KrigingResultModel",
     "KrigingRunner",
+    "KrigingTargetResult",
     "OrdinaryKriging",
     "SimpleKriging",
 ]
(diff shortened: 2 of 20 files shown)


packages/evo-compute/src/evo/compute/tasks/geostatistics/kriging.py
```diff
@@ -233,13 +402,54 @@ class KrigingParameters(BaseModel):
...
 
+    diagnostics: KrigingDiagnostics | None = Field(None, exclude=True)
+    """Optional diagnostics to write onto the target object alongside the estimate.
+
+    Only the diagnostics you select are computed. See :class:`KrigingDiagnostics`
+    for the available outputs and the shorthands each of them accepts.
+    """
+
+    @model_validator(mode="after")
+    def _validate_outputs(self) -> KrigingParameters:
```
aminabedi-bentley: Nit; Duplicate check keys on (operation, name_or_reference). A typed attribute named KV (→ update, ref …[?key=='kv']) plus kriging_efficiency="KV" (→ create "KV") is accepted, then both target the same attribute name. The SDK has `attr.name` at validation time and discards it — dedupe on the resolved name when known. 
Also two different expressions for the same attribute ([?name=='KV'] vs [?key=='kv']) slip through; the task's own `_validate_output_attribute_names` docstring warns this "would silently drop one of the requested outputs".   But I guess we don't want to validate further than what the task itself validates on the SDK level 🤔

packages/evo-compute/src/evo/compute/tasks/geostatistics/kriging.py
```diff
@@ -159,6 +177,148 @@ class BlockDiscretisation(BaseModel):
...
+}
+"""Recommended attribute name for each diagnostic, matching Leapfrog's naming convention."""
+
+
+def _object_identity(reference: str) -> str:
+    """The object a reference URL points at, ignoring any version query string."""
+    return reference.split("?", 1)[0]
+
+
+class KrigingDiagnostics(BaseModel):
```
aminabedi-bentley: Nit;
No attribute-type checking exists for KrigingDiagnostics fields. The schema constrains `valid` -> bool and `num_*` -> integer attribute types, but KrigingDiagnostics(valid=<Float64 attribute>) is accepted and fails server-side. The SDK already holds attribute_type so it _can_ validate the values against them. But unsure if it was in the scope of the ticket at all! Otherwise could be a separate ticket

mkdocs/site/packages/evo-compute/typed-objects/source-target/UpdateAttribute.html
```diff
@@ -39,7 +39,7 @@
                                 </a>
                             </li>
                             <li class="nav-item">
-                                <a rel="next" href="../../../evo-files/index.html" class="nav-link">
+                                <a rel="next" href="attribute_spec.html" class="nav-link">
```
aminabedi-bentley: Some of the files referenced in these mkdoc generated files do not even exist, I dug a bit into it an apparently it is an existing issue where `mkdocs-build.yaml` is including untracked files in its diff generation that it uses to generate the docs.
It is a non-issue at the moment as I don't see them being published anywhere, but I'd tag @GriffinBaxterSeequent for the unlikely case that this is not on their radar.

aminabedi-bentley (approved):
LGTM; with some minor comments.

packages/evo-compute/src/evo/compute/tasks/geostatistics/kriging.py
amirabbas-jalali-bentley: Good catch! fixed

packages/evo-compute/src/evo/compute/tasks/geostatistics/kriging.py
amirabbas-jalali-bentley: Good point, added!

mkdocs/site/packages/evo-compute/typed-objects/source-target/UpdateAttribute.html
GriffinBaxterSeequent: This is a non-issue as you say, but thanks for the heads up. The next/prev links aren't used at all when propagated to the developer portal.

GriffinBaxterSeequent (approved):
Happy with this from a evo-python-sdk-maintainers perspective, cheers!

packages/evo-compute/src/evo/compute/tasks/geostatistics/kriging.py
```diff
@@ -233,13 +431,68 @@ class KrigingParameters(BaseModel):
...
+                keys = [("reference", spec.reference)]
+            else:
+                keys = [("name", known_name), ("reference", spec.reference)]
+            for key in keys:
+                if (owner := written_by.get(key)) is not None:
+                    raise ValueError(
+                        f"{owner} and {label} both write to the attribute {key[1]!r}. "
+                        "Every kriging output needs its own attribute."
+                    )
+                written_by[key] = label
```
PaulCaygill-Seequent: Nit: This isn't super readable; I spent a lot of time trying to understand what we were checking for. 

As I read it, it looks like it raises a value error if you've got duplicate "references" or "names", is that right?

packages/evo-compute/tests/geostatistics/test_kriging_tasks.py
```diff
@@ -577,6 +586,356 @@ def test_kriging_params_without_filter(self):
...
+
+    def test_a_pending_attribute_is_not_checked(self):
+        """A new attribute is created with the diagnostic's type, so there is nothing to check."""
+        diagnostics = KrigingDiagnostics(valid=_create_pending_attribute("valid"))
+
+        self.assertEqual(diagnostics.valid, CreateAttribute(name="valid"))
+
+
+class TestKrigingDiagnosticOutputValidation(TestCase):
+    """Tests that every kriging output writes to its own attribute on the target object."""
```
PaulCaygill-Seequent: have you tested the duplication validation for output keys in the method `_validate_outputs` ?

PaulCaygill-Seequent (approved):
Small things, around the same area of code around Diagnostic validation, otherwise nothing jumping out as an issue

packages/evo-compute/src/evo/compute/tasks/geostatistics/kriging.py
amirabbas-jalali-bentley: Yes, you are right, I will try to make it more readable

packages/evo-compute/tests/geostatistics/test_kriging_tasks.py
amirabbas-jalali-bentley: Yes, `TestKrigingDiagnosticOutputValidation` covers it. I will add two more tests to make sure it doesn't reject combinations it shouldn't.

aminabedi-bentley (approved):

PR #350, opened 2026-09-14
Remove API preview from IMS users endpoint

keerynj04:
## Description
- Updated the IMS users endpoint to be out of preview
- Added models for inline workspace variables

## Checklist


packages/evo-sdk-common/src/evo/workspaces/endpoints/api/folders_api.py
```diff
@@ -1,743 +0,0 @@
-#  Copyright © 2025 Bentley Systems, Incorporated
```
samiulseequent, line 1: Why we are deleting this one?

packages/evo-sdk-common/src/evo/workspaces/endpoints/api/groups_api.py
```diff
@@ -55,14 +55,86 @@ class GroupsApi:
     def __init__(self, connector: APIConnector):
         self.connector = connector
 
-    async def get_ims_users(
+    async def get_group_description(
```
samiulseequent, line 58: Is this added here because we missed it generating this part in the past?

packages/evo-sdk-common/src/evo/workspaces/endpoints/api/admin_api.py
```diff
@@ -385,7 +385,7 @@ async def get_workspace_admin(
         deleted: bool | None = None,
         additional_headers: dict[str, str] | None = None,
         request_timeout: int | float | tuple[int | float, int | float] | None = None,
-    ) -> WorkspaceRoleOptionalResponse:  # noqa: F405
+    ) -> WorkspaceAdminResponse:  # noqa: F405
```
samiulseequent, line 388: You manually updated it, or its autogenerated?

packages/evo-sdk-common/src/evo/workspaces/endpoints/api/workspaces_api.py
```diff
@@ -672,8 +672,8 @@ async def list_workspaces_summary(
         order_by: str | None = None,
         filter_created_by: str | None = None,
         created_by: str | None = None,
-        created_at: str | None = None,
-        updated_at: str | None = None,
+        created_at: CreatedAtFilter | None = None,  # noqa: F405
```
samiulseequent, line 675: Why this one needs to change? others are still "str" as type?

packages/evo-sdk-common/src/evo/workspaces/endpoints/models.py
```diff
@@ -567,43 +593,103 @@ class WorkspaceRoleRequiredResponse(CustomBaseModel):
...
+    root: Annotated[
+        WorkspaceRoleRequiredResponse | WorkspaceRoleRequiredResponseWithThumbnailLink,
+        Field(title="WorkspaceResponse"),
+    ]
+    """
+    Named union of create/update workspace responses. Shared component schema for inline anyOf references.
+    """
+
+
+class DiscoveryResponse(CustomBaseModel):
```
samiulseequent, line 691: Why this one is added here? Is it something we missed in the past to generate?

samiulseequent:
@keerynj04 I think we need to update the evo-sdk-common for the automation tests in the workspace repo, right? I see under the post-deployment tests we have a test file called "tests_public_sdk.py".

packages/evo-sdk-common/src/evo/workspaces/endpoints/api/workspaces_api.py
keerynj04: `created_at` and `updated_at` both accept `str | list[str]` for date range filtering now. The class is so the both SDK generators correctly see the class and generate the code correctly, so this combined with the other fix in workspace's OpenAPI generation makes them autogenerate correctly. 

It just hadn't been re-generated since that change to what those parameters accepted so it only came up now.

packages/evo-sdk-common/src/evo/workspaces/endpoints/models.py
keerynj04: Yeah missed in the past

packages/evo-sdk-common/src/evo/workspaces/endpoints/api/groups_api.py
keerynj04: Yep missed in the past

packages/evo-sdk-common/src/evo/workspaces/endpoints/api/admin_api.py
keerynj04: Autogenerated. This was one of the combined classes we needed because the generators weren't seeing the same classes so one was generating code that the other didn't have. Check the schema in workspace-service for that class definition.

packages/evo-sdk-common/src/evo/workspaces/endpoints/api/folders_api.py
keerynj04: Seems like this was moved out of our schema to a separate folders schema, so the regeneration removed it. I also tried regenerating with that new schema and didn't see any changes in evo-sdk-common, so I believe we are good to remove this?

samiulseequent (approved):

samiulseequent:
Pls update the version for the evo-sdk-common in project.toml.

PR #352, opened 2026-09-21
Blockmodels reports new result schema

AndreLobatoSeequent:
## Description

Updates the block model reporting models to the new report/comparison result schema. Previously each report cell was a bare number; it is now a `ReportCell` object carrying a numeric `value` (nullable) together with a `status` explaining why a value may be missing (e.g. all inputs ignored, or marked invalid).

The change is backwards compatible: the SDK parses both the new object schema and the legacy bare-number schema, so a single SDK version works against both older and newer Block Model Service deployments. The typed `ReportResult.to_dataframe()` helper keeps its existing numeric output by default, so existing user code is unaffected.

## Acceptance Criteria

- A new `ReportCell` model (`value: int | float | None`, `status: ReportCellStatus`) and a `ReportCellStatus` enum (`OK`, `PERCENT_CHANGE_FROM_ZERO`, `NO_DATA`, `INVALID`) are available.
- Report results (`ReportRow.values`) and comparison results (`ReportComparisonValue.from_value`, `to_value`, `difference`, `percent`) are expressed as `ReportCell` objects.
- Responses using the new schema are parsed correctly, including cells with a `null` value and a non-`OK` status.
- Responses using the legacy schema (bare numbers, and `null` comparison percents) continue to parse, with each value upgraded to a `ReportCell` with `OK` status.
- `ReportResult.to_dataframe()` remains backwards compatible, emitting the numeric value by default, with `None`/`NaN` for cells that could not be aggregated.
- Cell status is available to callers without breaking the default output.
- `ReportCell` and `ReportCellStatus` are exported from the typed package.

## Implementation Notes

- **`endpoints/models.py`**
  - Added `ReportCellStatus` enum and `ReportCell` model.
  - Changed `ReportRow.values` to `list[ReportCell]` and the four `ReportComparisonValue` fields to `ReportCell`.
  - Added a `model_validator(mode="before")` on `ReportCell` that upgrades a bare number or `null` (legacy schema) into `{ "value": <n>, "status": "OK" }`. This keeps parsing compatible with both schemas. Legacy data flattened invalid/ignored cells before storage, so upgraded cells are marked `OK` (no per-cell status was available to recover).
- **`typed/report.py`**
  - `ReportResult.to_dataframe()` now unpacks each cell. By default only the numeric `value` is placed in the frame (unchanged shape); a new `include_status=True` argument adds a companion `"{column} status"` column.
  - Added a small `_unpack_cell` helper that handles both the object and legacy bare-number forms.
  - Exported `ReportCell` and `ReportCellStatus` from the typed package.
- The change is a non-breaking, additive update (minor version bump).

## Test Cases

Added to `tests/test_typed_report.py`:

- **`to_dataframe` (legacy)** — bare-number rows still produce the expected numeric DataFrame.
- **`to_dataframe` (new schema)** — `OK`/`NO_DATA`/`INVALID` cells; numeric value by default, with `NaN` for null/invalid cells and no status column unless requested.
- **`to_dataframe` with `include_status=True`** — adds a per-column status column with the correct status values.
- **`ReportCell` parsing** — new object schema (value + status), legacy bare number, and legacy `null` all validate correctly.
- **`ReportRow` parsing** — accepts both a list of `ReportCell` objects and a legacy list of bare numbers.
- **`ReportComparisonValue` parsing** — new schema (including a `PERCENT_CHANGE_FROM_ZERO` null percent) and the legacy bare-number schema (including a `null` percent) both parse.

## Checklist


packages/evo-blockmodels/src/evo/blockmodels/endpoints/models.py
```diff
@@ -557,6 +558,48 @@ class ReportColumn(CustomBaseModel):
...
+    PERCENT_CHANGE_FROM_ZERO = "PERCENT_CHANGE_FROM_ZERO"
+    NO_DATA = "NO_DATA"
+    INVALID = "INVALID"
+
+
+class ReportCell(CustomBaseModel):
+    """A single report/comparison cell: a numeric value plus a status.
+
+    The status explains why a value may be `null` (could not be aggregated/calculated).
+    See `ReportCellStatus` for the possible reasons.
```
cameron-stevenson-seequent: suggestion: the BMS "change schema" PR now has comments about each status, you may want to update `ReportCellStatus` here to have the same comments too.

https://github.com/seequent/block-model-service/pull/3440/changes#diff-1959642496206a02fb92ed7e0fedd9b59e1098c24bca06a115ff11464848b7cdR40-R46

packages/evo-blockmodels/src/evo/blockmodels/typed/report.py
```diff
@@ -247,6 +251,18 @@ class ReportSpecificationData:
...
+def _unpack_cell(cell: Any) -> tuple[int | float | None, str | None]:
+    """Return the ``(value, status)`` pair for a report cell.
+
+    Handles both the current schema, where a cell is a ``{"value", "status"}``
+    object, and the legacy schema, where a cell was a bare number.
+    """
+    if isinstance(cell, dict):
+        return cell.get("value"), cell.get("status")
+    # Legacy schema: the cell was a plain number (or None).
+    return cell, None
```
cameron-stevenson-seequent: question: the [BMS upgrade behaviour](https://github.com/seequent/block-model-service/pull/3440/changes#diff-1959642496206a02fb92ed7e0fedd9b59e1098c24bca06a115ff11464848b7cdR97) is to give non-null values a status of "OK", should we be consistent with that here?

packages/evo-blockmodels/src/evo/blockmodels/typed/report.py
```diff
@@ -279,12 +295,23 @@ def _from_api_result(cls, result: APIReportResult) -> ReportResult:
...
         """Convert the report result to a pandas DataFrame.
 
         Returns a DataFrame with one row per category/cut-off combination,
         containing the aggregated values for each report column.
 
+        Each cell in the report is a `ReportCell` (`{"value": ..., "status": ...}`).
+        By default only the numeric ``value`` is placed in the DataFrame (with
+        ``None`` where a value could not be aggregated), keeping the output the
+        same shape as previous SDK versions. Set ``include_status=True`` to add a
+        companion ``"{column} status"`` column exposing each cell's status.
```
cameron-stevenson-seequent: suggestion: This can theoretically break if a user has themselves defined a report where two columns are named like "col_1" and "col_1 status". 

My head went towards pandas having some way of grouping columns, so I asked copilot for ideas. I think its third idea of using struct columns but only if `include_status` is true could be a good lead.

Image

packages/evo-blockmodels/tests/test_typed_report.py
```diff
@@ -242,3 +286,79 @@ def test_use_in_report_spec(self) -> None:
...
+        """A bare number (legacy schema) is upgraded to a ReportCell with OK status."""
+        cell = models.ReportCell.model_validate(2.5)
+        self.assertEqual(cell.value, 2.5)
+        self.assertEqual(cell.status, ReportCellStatus.OK)
+
+    def test_report_cell_legacy_null(self) -> None:
+        """A bare null (legacy comparison percent) is upgraded to a ReportCell."""
+        cell = models.ReportCell.model_validate(None)
+        self.assertIsNone(cell.value)
+        self.assertEqual(cell.status, ReportCellStatus.OK)
```
cameron-stevenson-seequent: issue: I believe in older versions of BMS, the only thing producing a value of null was for `percent`, when the `from_value` is 0. So I think the status should be `PERCENT_CHANGE_FROM_ZERO` in this case.

packages/evo-blockmodels/tests/test_typed_report.py
```diff
@@ -242,3 +286,79 @@ def test_use_in_report_spec(self) -> None:
...
+
+    def test_comparison_value_legacy_schema(self) -> None:
+        """ReportComparisonValue accepts the legacy bare-number schema (percent may be null)."""
+        value = models.ReportComparisonValue.model_validate(
+            {"from_value": 0.0, "to_value": 3.0, "difference": 3.0, "percent": None}
+        )
+        self.assertEqual(value.from_value.value, 0.0)
+        self.assertEqual(value.from_value.status, ReportCellStatus.OK)
+        self.assertIsNone(value.percent.value)
+        self.assertEqual(value.percent.status, ReportCellStatus.OK)
```
cameron-stevenson-seequent: As above, `PERCENT_CHANGE_FROM_ZERO`?

cameron-stevenson-seequent (commented):
Good start on an approach that accepts the breaking change from the API but doesn't make a breaking change on the SDK.

I'm also curious whether it would be any easier/harder to allow the SDK change to be breaking as well? Given our queries on reporting usage indicated no significant usage so far.

PR #353, opened 2026-09-22
Loosen ObjectReference hub URL schema validation

rohancrookbain-seequent:
## Description

Widens the validation on `ObjectReference`, so that non-HTTPS schemas are allowed.
It doesn't really make sense to validate this there, since in all likelihood the URL comes from an (already validated) `Environment`.

## Checklist


packages/evo-objects/src/evo/objects/data.py
```diff
@@ -91,8 +91,8 @@ def __new__(cls, value: str) -> ObjectReference:
         inst = str.__new__(cls, value)
 
         parsed = urlparse(value)
-        if parsed.scheme != "https":
-            raise ValueError("Reference must be a valid HTTPS URL")
+        if parsed.scheme not in ("http", "https"):
+            raise ValueError("Reference must be a valid HTTP or HTTPS URL")
```
rohancrookbain-seequent:
```suggestion
        if not parsed.scheme:
            # Valid hub URLs are always HTTPS, but it is sometimes useful to pass HTTP ones for testing, so these are allowed here too.
            raise ValueError("Reference must be a valid HTTPS URL")
```

Bill-Johnson-Seequent (approved):
Please also bump [the version number](https://github.com/SeequentEvo/evo-python-sdk/blob/main/packages/evo-objects/pyproject.toml#L4) - with that recommendation, this change is now ok to merge.

Bill-Johnson-Seequent (approved):

RyanMillerSeequent (approved):

PR #354, opened 2026-09-23
Fix tail extrapolation payloads for conditioned simulation and simulation report

amirabbas-jalali-bentley:
## Description

The conditioned-simulator and simulation-report tasks declare two fields in their `distribution` parameter as required but nullable, with no default:

- `distribution.tail_extrapolation`
- `distribution.tail_extrapolation.lower`

So the key has to be present, but `null` is a valid value for it. `TaskRunner` serialises parameters with `exclude_none=True`, so whenever either field was left unset the key was dropped and the task rejected the job:

- a distribution with weights but no tail extrapolation failed with `'tail_extrapolation' is a required property`
- a tail extrapolation with only an upper tail failed with `'lower' is a required property`

There was no way around it from the calling side, since passing `None` explicitly makes no difference to `exclude_none`.

### Changes

- `DistributionParams`, `SimulationReportDistribution` and `TailExtrapolationParams` now always emit these keys, as `null` when unset. This is the same `@model_serializer(mode="wrap")` approach `DeclusteringParameters` already uses to keep `power: null`. Every other field is still omitted when it's `None`, and no other task is affected.
- `SimulationReportDistribution.tail_extrapolation` was annotated as `None`, so a real tail extrapolation couldn't be passed to the simulation report at all. It now takes the same `TailExtrapolationParams` as the conditioned simulator, since both tasks accept the same shape.

I considered switching `TaskRunner` to `exclude_unset=True` instead, but it doesn't help on its own (a field the caller never set is still dropped), and it would stop sending the client-side defaults several tasks rely on, such as the kriging `method` and the conditioned simulator's defaults.

No breaking changes. Payloads that were already accepted are unchanged; only the ones that used to be rejected now carry the extra `null` keys. `lower` stays optional.

### Testing

- New tests pin the explicit `null`s on the wire for both tasks, check that a real `lower` isn't overwritten, and check that the simulation report takes a real tail extrapolation. Four of the five fail without the fix.
- Validated the resulting payloads against both tasks' JSON schemas: no distribution, a distribution without tail extrapolation, an upper tail only, and both tails are all accepted.
- `evo-compute` tests: 418 passed. `ruff check` and `ruff format --check` are clean.

## Checklist


higordearaujo-bsy (approved):

GriffinBaxterSeequent (approved):
All good from a `evo-python-sdk-maintainers` perspective!

